PR #6, opened 2017-01-04
Remove fromEvent, from. Export fromArray, fromIterable

briancavalier:
Remove `fromEvent`.  We'll direct people to [`@most/dom-event`](https://github.com/mostjs/dom-event), and the upcoming event-emitter add-on.  Remove `from` to get rid of runtime type inspection-based dispatch.  Export `fromArray` and `fromIterable` explicitly.

A few other things along the way:

* Rename internal `cons` to `startWith`
* Rename internal `of` to `just`
* convert touched files to es6 (didn't convert files that were auto-refactored by Webstorm due to `of` -> `just` rename)


davidchase:
Looks good and snuck a little compose :smile:

PR #8, opened 2017-01-05
WIP: Add runEffects(), remove defaultScheduler

briancavalier:
⚠️ DO NOT MERGE ⚠️ 

WIP: Add runEffects() with the goal of replacing the defaultScheduler instance with a createDefaultScheduler() function, and removing observe, drain, and reduce which invisibly use defaultScheduler

### Todo

- [x] Remove reduce, observe, drain exports
- [x] Remove defaultScheduler
- [x] Add createDefaultScheduler()


briancavalier:
Removing reduce, observe, and drain *entirely* means converting lots more unit tests.  Even though it's relatively mechanical, it's proving to be quite time consuming.  So, instead of letting that be a blocker to releasing an initial version that we can start to use and get some experience with, we might be able to rely on tree shaking to publish something sooner.

If we remove reduce, observe, and drain from `src/index.js` exports.  That should allow them to be culled from the build, while keeping them available for the unit tests that currently rely on them.  Then, we could publish a 0.x build, and continue converting unit tests.

Thoughts?

PR #11, opened 2017-01-06
Port sample from @most/sample

### LINKED ISSUE #9: What to do with sample, sampleWith?
Since `@most/sample` exists as a "better" sampling implementation, what should we do with the existing `sample` and `sampleWith` functions in `@most/core`?

Some possible options:

1. Keep them as is.  I'm not sure why we would do this.  Are there any compelling reasons?
1. Keep them as exports, but reimplement them using `@most/sample`.  I can't think of a compelling reason to do this, since `@most/core` has no backward compatibility concerns.
1. Remove them and direct people to `@most/sample`
1. Deprecate `@most/sample`, and integrate its API and implementation directly into `@most/core`.  If we did this, we'd still need to decide whether to keep the existing `sample` and `sampleWith`.  If we kept them, I can't see a reason *not* to reimplement them.

I don't have a sense for how often they are used, which may be a factor: if they're common, then people would always find themselves needing both `@most/core` and `@most/sample` that's an interesting data point that could push us in the direction of 4 rather than 3.

Thoughts?

briancavalier:
Close #9 

Port sample from @most/sample.  Remove older sample API and tests.

### Todo

- [x] Update index.js to export new sample and sampleArray
- [ ] ~~sampleArray unit tests (currently failing test as reminder)~~. In spirit of stripping away first before adding things back, I decided to remove `sampleArray`, since it can be made trivially from `sample` + `combineArray`

### DIFF
--- a/src/combinator/sample.js
+++ b/src/combinator/sample.js
@@ -2,17 +2,10 @@
 
 import Stream from '../Stream'
 import Pipe from '../sink/Pipe'
-import { combineArray } from './combine'
 import * as dispose from '../disposable/dispose'
-import { curry3 } from '@most/prelude'
 
-export const sample = curry3((f, sampler, stream) =>
-  new Stream(new SampleSource(f, sampler, stream)))
-
-const arrayId = (...values) => values
-
-export const sampleArray = curry3((f, sampler, arrayOfStreams) =>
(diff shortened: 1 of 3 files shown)


TylorS:
:shipit:

PR #15, opened 2017-01-10
Update and simplify disposables

briancavalier:
Mostly just cleanup and conversion to ES.

### DIFF
--- a/src/disposable/dispose.js
+++ b/src/disposable/dispose.js
@@ -2,6 +2,8 @@
 
 import Disposable from './Disposable'
 import SettableDisposable from './SettableDisposable'
+import MemoizedDisposable from './MemoizedDisposable'
+import { disposeSafely } from './disposeSafely'
 import { isPromise } from '../Promise'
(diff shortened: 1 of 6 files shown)


src/disposable/dispose.js
```diff
@@ -72,47 +62,46 @@ function disposeSafely (disposable) {
...
  * Wrap an existing disposable (which may not already have been once()d)
  * so that it will only dispose its underlying resource at most once.
  * @param {{ dispose: function() }} disposable
  * @return {Disposable} wrapped disposable
  */
-export function once (disposable) {
-  return new Disposable(disposeMemoized, memoized(disposable))
-}
+export const once = disposable =>
+  new Disposable(disposeOne, new MemoizedDisposable(disposable))
```
briancavalier: Lol, this seems rather ridiculous after sleeping on it.  Probably can just be:

```js
export const once = disposable =>
  new MemoizedDisposable(disposable)

davidchase: simplify all the things! ✊

PR #20, opened 2017-01-12
chore(core) update dist

davidchase:
was trying to use this one the examples as discussed in the slack but half way thru realized the dist/build had not been updated


briancavalier:
Thanks!

PR #23, opened 2017-01-13
Add a simple README. Update package.json description and keywords

briancavalier:


README.md
```diff
@@ -35,10 +33,11 @@ First, the HTML fragment for the inputs and a place to display the live result:
 </form>
 ```
 
-Using most.js to make it reactive:
+Using `@most/core` to make it reactive:
 
 ```es6
-import { fromEvent, combine } from 'most'
+import { combine, newDefaultScheduler } from '@most/core'
+import { input } from '@most/dom-event'
```
davidchase: is this missing `runEffects` used below?

briancavalier: Oops, thanks!

README.md
```diff
@@ -36,7 +36,7 @@ First, the HTML fragment for the inputs and a place to display the live result:
 Using `@most/core` to make it reactive:
 
 ```es6
-import { combine, newDefaultScheduler } from '@most/core'
+import { combine, runEffects, newDefaultScheduler } from '@most/core'
```
davidchase: also forget to ask if its called [createDefaultScheduler](https://github.com/mostjs/core/blob/master/src/index.js#L178) rather than `newDefaultScheduler` ? 😄

briancavalier: I had gone with the naming from #21, but it probably makes sense to stick with the naming from master until #21 is merged.  I'll change it, thanks!

PR #26, opened 2017-01-13
chore(META): add basic configurations for northbrook

TylorS:
Adds a simple northbrook configuration to use eslint and rollup plugins.

AFFECTS: @most/core

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,10 +11,12 @@
     "dist"
   ],
   "scripts": {
+    "commit": "northbrook commit",
     "unit-test": "buster-test",
-    "test": "eslint src test && npm run unit-test",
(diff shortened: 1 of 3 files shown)


davidchase:
🚢 🇮🇹

package.json
```diff
@@ -11,10 +11,12 @@
     "dist"
   ],
   "scripts": {
+    "commit": "northbrook commit",
```
briancavalier: How about we pick *either* the name `nb` or `northbrook` and use it in all the scripts, rather than mixing--it's `northbrook` everywhere except the `"build"` script atm.  Any preference?

TylorS: I generally use `northbrook` in my scripts to make it clearer what I'm doing. Will make them all as such if you want?

briancavalier: :+1: Yep, either name is fine with me.  I just wanted to pick one.

TylorS:
Updated and rebased :boom:

PR #27, opened 2017-01-16
Update type-definitions

TylorS:
Updates the type definitions to be much more maintainable and correct for the new repo.


type-definitions/combinator/combine.d.ts
```diff
@@ -0,0 +1,7 @@
+import { Stream } from '../types';
+
+export function combine<A, B, R>(fn: (a: A, b: B) => R, a: Stream<A>, b: Stream<B>): Stream<R>;
+export function combine<A, B, C, R>(fn: (a: A, b: B, c: C) => R, a: Stream<A>, b: Stream<B>, c: Stream<C>): Stream<R>;
+export function combine<A, B, C, D, R>(fn: (a: A, b: B, c: C, d: D) => R, a: Stream<A>, b: Stream<B>, c: Stream<C>, d: Stream<D>): Stream<R>;
+export function combine<A, B, C, D, E, R>(fn: (a: A, b: B, c: C, d: D, e: E) => R,a: Stream<A>,b: Stream<B>, c: Stream<C>, d: Stream<D>, e: Stream<E>): Stream<R>;
```
briancavalier: Minor: coupla commas here that needs spaces after

type-definitions/combinator/combine.d.ts
```diff
@@ -0,0 +1,7 @@
...
+export function combine<R>(f: (...args: any[]) => R, ...streams: Stream<any>[]): Stream<R>;
```
briancavalier: I like that array shortcut syntax (`any[]`).  I don't think flow has that.

type-definitions/combinator/errors.d.ts
```diff
@@ -0,0 +1,6 @@
+import { Stream } from '../types';
+
+export function recoverWith<A, B>(p: (a: B) => Stream<A>, s: Stream<A>): Stream<A>;
```
briancavalier: Can TS constrain the bounds `B` to `Error`?  For example, [here's what recoverWith looks like in flow](https://github.com/mostjs/core/blob/master/src/index.js.flow#L242).  Also, might be a good visual cue to rename `B` -> `E` to indicate an error type.

type-definitions/combinator/errors.d.ts
```diff
@@ -0,0 +1,6 @@
...
+export function recoverWith<A, B>(p: (a: B) => Stream<A>): (s: Stream<A>) => Stream<A>;
+
+export function throwError(e: Error): Stream<any>;
```
briancavalier: The flow types parameterize and constrain the error type here.  I don't see a compelling reason for it, though, other than consistency since right now, the error type doesn't really help the caller.  I could go either way.

type-definitions/combinator/zip.d.ts
```diff
@@ -0,0 +1,7 @@
+import { Stream } from '../types';
+
+export function zip<A, B, R>(fn: (a: A, b: B) => R, a: Stream<A>, b: Stream<B>): Stream<R>;
+export function zip<A, B, C, R>(fn: (a: A, b: B, c: C) => R, a: Stream<A>, b: Stream<B>, c: Stream<C>): Stream<R>;
+export function zip<A, B, C, D, R>( fn: (a: A, b: B, c: C, d: D) => R, a: Stream<A>, b: Stream<B>, c: Stream<C>, d: Stream<D>): Stream<R>;
+export function zip<A, B, C, D, E, R>(fn: (a: A, b: B, c: C, d: D, e: E) => R, a: Stream<A>, b: Stream<B>, c: Stream<C>, d: Stream<D>, e: Stream<E>): Stream<R>;
+export function zip<R>(f: (...args: any[]) => R, ...streams: Stream<any>[]);
```
briancavalier: Need zipArray

type-definitions/source/fromArray.d.ts
```diff
@@ -0,0 +1,4 @@
+import { Stream } from '../types';
+
+export function fromArray<A>(array: Array<A> | ArrayLike<A>): Stream<A>;
```
briancavalier: Interesting.  Where does the ArrayLike type come from?

type-definitions/source/fromIterable.d.ts
```diff
@@ -0,0 +1,4 @@
+import { Stream } from '../types';
+
+export function fromIterable<A>(iterable: Iterable<A>): Stream<A>;
+export function fromIterable(iterable: Iterable<any>): Stream<any>;
```
briancavalier: Why the `any` variant?

type-definitions/source/generate.d.ts
```diff
@@ -0,0 +1,3 @@
+import { Stream } from '../types';
+
+export function generate<A>(g: GeneratorFunction, ...args: Array<any>): Stream<A>;
```
briancavalier: Oops!  I think `generate` is missing from the flow type defs!

type-definitions/source/iterate.d.ts
```diff
@@ -0,0 +1,4 @@
+import { Stream } from '../types';
+
+export function iterate<A>(f: (a: A) => A | Promise<A>, a: A): Stream<A>;
+export function iterate<A>(f: (a: A) => A | Promise<A>, a: A): Stream<A>;
```
briancavalier: One of these needs to have its argument list split 😄

type-definitions/source/unfold.d.ts
```diff
@@ -0,0 +1,3 @@
+import { Stream, SeedValue } from '../types';
+
+export function unfold<A, B, S>(f: (seed: S) => SeedValue<S, B|Promise<B>>, seed: S): Stream<B>;
```
briancavalier: Need curried variants

type-definitions/scheduler.d.ts
```diff
@@ -0,0 +1,8 @@
+import { Timer, Timeline, Scheduler } from './types';
+
```
briancavalier: Need [newScheduler](https://github.com/mostjs/core/blob/master/src/index.js#L183)

type-definitions/PropagateTask.d.ts
```diff
@@ -0,0 +1,15 @@
+import { Sink, Task } from './types';
+
+export function propagateTask<T>(time: number, value: T, sink: Sink<T>): Task;
```
briancavalier: propagateTask takes a run function as its first argument.  The run function takes 4 args.  [Here's the flow defs for it](https://github.com/mostjs/core/blob/master/src/index.js.flow#L254-L265).  Since one of the 4 run function args is the actual PropagateTask instance itself, and it has the `active` field, which is useful (and used in a few places already), you may end up needing to have a specific PropagateTask type (I had to do that in the flow defs).

type-definitions/combinator/zip.d.ts
TylorS: Fixed!

type-definitions/source/fromArray.d.ts
TylorS: TypeScript's "standard library"

type-definitions/source/generate.d.ts
TylorS: Nice catch!

type-definitions/source/unfold.d.ts
TylorS: Fixed!

type-definitions/scheduler.d.ts
TylorS: Fixed!

TylorS:
Just pinging to see if there is anything else to do here?

briancavalier:
Thanks for the awesome work here, @TylorS.

PR #29, opened 2017-01-17
fix(core): Fix newDefaultScheduler argument order

briancavalier:
AFFECTS: @most/core

🤦‍♂️ Enforcing index's own types upon itself would have prevented this.  I'll see if that is easy to do.


davidchase:
:shipit:

PR #30, opened 2017-01-18
feat(core): Add task arg to PropagateTask run function

### LINKED ISSUE #19: Remove the need to call PropagateTask's run function with `this`
PropagateTask currently calls its run function [with itself as `this`](https://github.com/cujojs/most/blob/master/src/scheduler/PropagateTask.js#L34).  This was an easy (but not simple) way to provide access to the `active` state, which is needed to stop early in the case where the task is canceled due to any of the various ways the stream could be disposed, such as `take()`.

That leads to WTFs like [where does `this.active` come from?!?](https://github.com/cujojs/most/blob/master/src/source/fromArray.js#L21).  It would be nice to find a better solution.

briancavalier:
Remove the need to use thisArg in PropagateTask run functions by passing the task as an additional arg.  This also moves PropagateTask helpers from static methods to plain functions and switches PropagateTask to use named exports.  Convert touched sources to ES2015.

Close #19

AFFECTS: @most/core

ISSUES CLOSED: #19

### DIFF
--- a/src/combinator/delay.js
+++ b/src/combinator/delay.js
@@ -5,7 +5,7 @@
 import Stream from '../Stream'
 import Pipe from '../sink/Pipe'
 import * as dispose from '../disposable/dispose'
-import { propagateEvent, propagateEnd } from '../scheduler/PropagateTask'
+import { propagateEventTask, propagateEndTask } from '../scheduler/PropagateTask'
 
 /**
  * @param {Number} delayTime milliseconds to delay each item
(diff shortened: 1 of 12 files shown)


briancavalier:
Hmmm, seeing [what I had to do to the exports in src/index](https://github.com/mostjs/core/pull/30/commits/f02b9d964ae9d6b5f34f05690f6995b64a749e4f), I'm not sure switching from static methods to exported functions is the right thing.  It adds functions directly to the public API namespace, rather than nesting them under the PropagateTask constructor.

Which seems better?

TylorS:
Code seems good :+1: 

> Which seems better?

I'm in favor of using the functions, but one tough thing is the names, they're not quite as obvious as the static constructors.

Also if they're going into the public API (I have no issue with that), should they also be curried?

briancavalier:
> they're not quite as obvious as the static constructors

Agreed.  Any suggestions?  `propagateEventTask` et. al.?

> should they also be curried?

Good catch.  Indeed they should be.  Once we make a final decision on statics vs functions and names, I'll add currying, too.

TylorS:
I like `propagateEventTask` :+1:

davidchase:
i like the names as well its they're bit long but def worth it 😄  and i also dont mind exporting functions over static

TylorS:
:shipit:

PR #33, opened 2017-01-25
chore(test): Switch 1 existing mocha test to use @briancavalier/assert

briancavalier:
AFFECTS: @most/core


davidchase:
:shipit:

PR #36, opened 2017-01-27
test(disposable): convert sink & disposable tests

TylorS:
convert disposable tests to mocha + @briancavalier/assert
convert sink tests to mocha + @briancavalier/assert

AFFECTS: @most/core

### DIFF
--- a/test/disposable/SettableDisposable-test.js
+++ b/test/disposable/SettableDisposable-test.js
@@ -1,5 +1,5 @@
 import { describe, it } from 'mocha'
-import { eq, throws } from '@briancavalier/assert'
+import { throws, is } from '@briancavalier/assert'
 
 import { default as SettableDisposable } from '../../src/disposable/SettableDisposable'
 
@@ -14,7 +14,7 @@ describe('SettableDisposable', function () {
 
     var x = d.dispose()
 
-    eq(x, data)
+    is(data, x)
   })
 
   it('should allow setDisposable after dispose', function () {
@@ -27,7 +27,7 @@ describe('SettableDisposable', function () {
       dispose: function () { return data }
     })
 
-    return p.then(eq(data))
+    return p.then(is(data))
   })
 
   it('should allow setDisposable at most once', function () {
--- a/test/disposable/SettableDisposable-test.js
+++ b/test/disposable/SettableDisposable-test.js
@@ -1,5 +1,5 @@
-import { spec, expect } from 'buster'
-const { describe, it } = spec
+import { describe, it } from 'mocha'
+import { eq, throws } from '@briancavalier/assert'
 
 import { default as SettableDisposable } from '../../src/disposable/SettableDisposable'
 
@@ -14,7 +14,7 @@ describe('SettableDisposable', function () {
 
     var x = d.dispose()
 
-    expect(x).toBe(data)
+    eq(x, data)
   })
 
   it('should allow setDisposable after dispose', function () {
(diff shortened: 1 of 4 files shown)


davidchase:
👍

test/disposable/SettableDisposable-test.js
```diff
@@ -14,7 +14,7 @@ describe('SettableDisposable', function () {
 
     var x = d.dispose()
 
-    expect(x).toBe(data)
+    eq(x, data)
```
briancavalier: Unfortunately, it's not obvious from the [eq docs](https://github.com/briancavalier/assert#eq--a--a--a), but the intent is for the parameter ordering to be `expected, actual`.  It's more obvious when you partially apply it in assertions like: `promise.then(eq(expected))`.  Not a huge deal, but I'd like to go with `expected, actual` wherever possible.

I feel like `is` may be the right assertion here.  For example, this test could pass if the implementation of `SettableDisposable.prototype.dispose` were changed to simple `return {}`, since `eq({}, {}) === true`.  IOW, it is possible for `d.dispose` to conjure up a value that will pass the test, but it's not possible for it to conjure up a reference that will *unless* it's returned by the wrapped disposable.  Does that make sense?

test/disposable/dispose-test.js
```diff
@@ -58,16 +58,15 @@ describe('tryDispose', function () {
...
   })
 
   it('should return disposable result promise', function () {
     var x = {}
     var spy = disposableSpy(returns, Promise.resolve(x))
 
-    return dispose.tryDispose(0, spy, failSink).then(function (y) {
-      expect(y).toBe(x)
-    })
+    return dispose.tryDispose(0, spy, failSink).then(eq(x))
```
briancavalier: Many of these tests were designed to verify the object identity, so that the function under test can't invent an equivalent value and pass the test.  I'm not really convinced it matters (and in a truly immutable environment, it wouldn't!), but wanted to see what you think.  I can see a couple options:

1. Use `is`
2. Use `eq`, but use a value that is unlikely ever to cause a false positive, like `Math.random()` instead of `{}`

Thoughts?

TylorS: Both seem good, but `is` is probably the more readily understood change.

test/sink/SafeSink-test.js
```diff
@@ -130,13 +127,13 @@ describe('SafeSink', function () {
...
-    expect(errorCalled).toBe(2)
+    eq(errorCalled, 2)
   })
 
   it('disable should return original sink', function () {
     var original = testSink(fail, fail, fail)
     var sink = new SafeSink(original)
 
-    expect(sink.disable()).toBe(original)
+    eq(sink.disable(), original)
```
briancavalier: Hmm, this one is interesting.  At first, I thought it was a compelling case to use `is`: `original` could be any sink, and (unfortunately) since sinks tend to hold mutable state, it seems important to ensure `sink.disable()` returns a reference to `original`.

But using `is` may imply that it's ok for `SafeSink` to mutate fields of `original` in some way.  If it did, then `is` would pass.  Come to think of it, `eq` would pass as well!

Oddly enough, it seems like the most strict test would be something like:

```js
eq(sink.disable(), testSink(fail, fail, fail))
```

Is that crazy?  Maybe unnecessarily paranoid ... `eq` is probably fine :)  but wanted to see what you think.

test/disposable/SettableDisposable-test.js
```diff
@@ -14,7 +14,7 @@ describe('SettableDisposable', function () {
 
     var x = d.dispose()
 
-    expect(x).toBe(data)
+    is(x, data)
```
briancavalier: I think `data` is expected and `x` is actual here.

PR #42, opened 2017-02-11
test(combinator) convert combinator tests

### LINKED ISSUE #37: Convert combinator/ tests
i will give this one a whirl (tonight/tomorrow am) if you guys dont mind so i can be more familiar with both nb and the assert lib

davidchase:
convert combinator tests to mocha + @briancavalier/assert

AFFECTS: @most/core

closes #37

### DIFF
--- a/packages/core/test/combinator/applicative-test.js
+++ b/packages/core/test/combinator/applicative-test.js
@@ -2,52 +2,46 @@ import { describe, it } from 'mocha'
 
 import { assertSame } from '../helper/stream-helper'
 import { ap } from '../../src/combinator/applicative'
-import { just as streamOf } from '../../src/source/core'
+import { just } from '../../src/source/core'
 
 const sentinel = { value: 'sentinel' }
 
 describe('ap', function () {
   it('should satisfy identity', function () {
     // P.of(function(a) { return a; }).ap(v) ~= v
-    const v = streamOf(sentinel)
-    return assertSame(ap(streamOf(function (x) { return x }), v), v)
+    const v = just(sentinel)
+    return assertSame(ap(just(x => x), v), v)
   })
 
   it('should satisfy composition', function () {
     // P.of(function(f) { return function(g) { return function(x) { return f(g(x))}; }; }).ap(u).ap(v
-    const u = streamOf(function (x) { return 'u' + x })
-    const v = streamOf(function (x) { return 'v' + x })
-    const w = streamOf('w')
+    const u = just(x => 'u' + x)
+    const v = just(x => 'v' + x)
+    const w = just('w')
 
     return assertSame(
-      ap(ap(ap(streamOf(function (f) {
-        return function (g) {
-          return function (x) {
-            return f(g(x))
-          }
-        }
-      }), u), v), w),
+      ap(ap(ap(just(f => g => x => f(g(x))), u), v), w),
       ap(u, ap(v, w))
     )
   })
 
   it('should satisfy homomorphism', function () {
     // P.of(f).ap(P.of(x)) ~= P.of(f(x)) (homomorphism)
-    function f (x) { return x + 'f' }
+    const f = x => x + 'f'
     const x = 'x'
-    return assertSame(ap(streamOf(f), streamOf(x)), streamOf(f(x)))
+    return assertSame(ap(just(f), just(x)), just(f(x)))
   })
 
   it('should satisfy interchange', function () {
     // u.ap(a.of(y)) ~= a.of(function(f) { return f(y); }).ap(u)
-    function f (x) { return x + 'f' }
+    const f = x => x + 'f'
 
-    const u = streamOf(f)
+    const u = just(f)
     const y = 'y'
 
     return assertSame(
-      ap(u, streamOf(y)),
-      ap(streamOf(function (f) { return f(y) }), u)
+      ap(u, just(y)),
+      ap(just(f => f(y)), u)
     )
   })
 })
--- a/packages/core/test/combinator/applicative-test.js
+++ b/packages/core/test/combinator/applicative-test.js
@@ -1,24 +1,23 @@
-import { spec } from 'buster'
-const { describe, it } = spec
+import { describe, it } from 'mocha'
 
 import { assertSame } from '../helper/stream-helper'
 import { ap } from '../../src/combinator/applicative'
 import { just as streamOf } from '../../src/source/core'
 
-var sentinel = { value: 'sentinel' }
+const sentinel = { value: 'sentinel' }
 
 describe('ap', function () {
   it('should satisfy identity', function () {
     // P.of(function(a) { return a; }).ap(v) ~= v
-    var v = streamOf(sentinel)
+    const v = streamOf(sentinel)
     return assertSame(ap(streamOf(function (x) { return x }), v), v)
   })
 
   it('should satisfy composition', function () {
     // P.of(function(f) { return function(g) { return function(x) { return f(g(x))}; }; }).ap(u).ap(v
-    var u = streamOf(function (x) { return 'u' + x })
-    var v = streamOf(function (x) { return 'v' + x })
-    var w = streamOf('w')
+    const u = streamOf(function (x) { return 'u' + x })
+    const v = streamOf(function (x) { return 'v' + x })
+    const w = streamOf('w')
 
     return assertSame(
       ap(ap(ap(streamOf(function (f) {
@@ -35,7 +34,7 @@ describe('ap', function () {
   it('should satisfy homomorphism', function () {
     // P.of(f).ap(P.of(x)) ~= P.of(f(x)) (homomorphism)
     function f (x) { return x + 'f' }
-    var x = 'x'
+    const x = 'x'
     return assertSame(ap(streamOf(f), streamOf(x)), streamOf(f(x)))
   })
 
@@ -43,8 +42,8 @@ describe('ap', function () {
     // u.ap(a.of(y)) ~= a.of(function(f) { return f(y); }).ap(u)
     function f (x) { return x + 'f' }
 
-    var u = streamOf(f)
-    var y = 'y'
+    const u = streamOf(f)
+    const y = 'y'
 
     return assertSame(
       ap(u, streamOf(y)),
--- a/packages/core/test/combinator/chain-test.js
+++ b/packages/core/test/combinator/chain-test.js
@@ -1,5 +1,5 @@
 import { describe, it } from 'mocha'
-import { assert, is, eq } from '@briancavalier/assert'
+import { assert, eq } from '@briancavalier/assert'
 import { spy } from 'sinon'
 
 import { chain, join } from '../../src/combinator/chain'
@@ -12,7 +12,7 @@ import { fromArray } from '../../src/source/fromArray'
 import Stream from '../../src/Stream'
 
 import { assertSame } from '../helper/stream-helper'
-import { ticks, collectEvents } from '../helper/testEnv'
+import { collectEventsFor } from '../helper/testEnv'
 import FakeDisposeSource from './../helper/FakeDisposeSource'
 
 const sentinel = { value: 'sentinel' }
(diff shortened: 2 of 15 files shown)


packages/core/test/combinator/chain-test.js
```diff
@@ -36,13 +36,13 @@ describe('chain', function () {
 
     return collectEvents(s, ticks(3))
       .then(events => {
-        assert.same(2, events.length)
+        is(2, events.length)
```
briancavalier: I've started to prefer writing assertions like this:

```js
// Array of (time, value) pairs representing the events we expect
const expected = [{ time: 1, value: 1 }, { time: 2, value: 2 }]

// collectEventsFor is a convenience function on top of collectEvents()
return collectEventsFor(3, s).then(eq(expected))
```

Do you guys like that style also?

packages/core/test/combinator/concatMap-test.js
```diff
@@ -34,85 +36,79 @@ describe('concatMap', function () {
...
   })
 
   it('should map lazily', function () {
-    var s1 = atTimes([{ time: 0, value: 0 }, { time: 1, value: 1 }])
+    const s1 = atTimes([{ time: 0, value: 0 }, { time: 1, value: 1 }])
 
-    var scheduler = ticks(4)
-    var s = concatMap.concatMap(function (x) {
+    const scheduler = ticks(4)
+    const s = concatMap.concatMap(function (x) {
```
briancavalier: Let's convert to arrow functions whenever possible.

packages/core/test/combinator/concatMap-test.js
```diff
@@ -34,85 +36,79 @@ describe('concatMap', function () {
...
+    const dispose = spy()
+    const inner = streamOf(sentinel)
+    const outer = streamOf(inner)
 
-    var s = concatMap.concatMap(identity, new Stream(new FakeDisposeSource(dispose, outer.source)))
+    const s = concatMap.concatMap(identity, new Stream(new FakeDisposeSource(dispose, outer.source)))
 
     return drain(s).then(function () {
-      expect(dispose).toHaveBeenCalled()
+      assert(dispose.called)
```
briancavalier: Arrow func

packages/core/test/combinator/concatMap-test.js
```diff
@@ -34,85 +36,79 @@ describe('concatMap', function () {
...
-    var inner = new Stream(new FakeDisposeSource(dispose, streamOf(sentinel).source))
+    const dispose = spy()
+    const inner = new Stream(new FakeDisposeSource(dispose, streamOf(sentinel).source))
 
-    var s = concatMap.concatMap(identity, streamOf(inner))
+    const s = concatMap.concatMap(identity, streamOf(inner))
 
     return drain(s).then(function () {
-      expect(dispose).toHaveBeenCalled()
+      assert(dispose.called)
```
briancavalier: Arrow func

packages/core/test/combinator/applicative-test.js
```diff
@@ -1,24 +1,23 @@
-import { spec } from 'buster'
-const { describe, it } = spec
+import { describe, it } from 'mocha'
 
 import { assertSame } from '../helper/stream-helper'
 import { ap } from '../../src/combinator/applicative'
 import { just as streamOf } from '../../src/source/core'
```
briancavalier: Not strictly necessary, but since you're in this code, how about renaming `streamOf` usages to `just` so we can get rid of this `as`?

packages/core/test/combinator/concatMap-test.js
```diff
@@ -34,85 +36,79 @@ describe('concatMap', function () {
...
-    var inners = values.map(function (x, i) {
+    const inners = values.map(function (x, i) {
       return new Stream(new FakeDisposeSource(spies[i], streamOf(x).source))
     })
 
-    var s = concatMap.concatMap(identity, fromArray(inners))
+    const s = concatMap.concatMap(identity, fromArray(inners))
 
     return drain(s).then(function () {
       spies.forEach(function (spy) {
```
briancavalier: Arrow funcs

packages/core/test/combinator/continueWith-test.js
```diff
@@ -10,6 +9,6 @@ describe('continueWith', () => {
   it('when f throws, should propagate error', () => {
     const error = new Error()
     const s = continueWith(() => { throw error }, just(0))
-    return drain(s).then(fail, e => assert.same(error, e))
+    return drain(s).then(rejects, eq(error))
```
briancavalier: The preferred way to use `rejects` here is:

```js
return rejects(drain(s)).then(is(error))
```

Another way would be to use `@briancavalier/assert`'s `fail`, similar to how it was with buster:

```js
return drain(s).then(fail, is(error))
```

I think I'd also prefer `is` here, although I don't feel too strongly about it.

packages/core/test/combinator/errors-test.js
```diff
@@ -41,20 +40,20 @@ describe('recoverWith', () => {
...
       throw new Error()
-    }, s).catch(e => assert.same(sentinel, e))
+    }, s).catch(eq(sentinel))
   })
 
   it('when f throws, should propagate error', () => {
     const error = new Error()
     const s = recoverWith(x => { throw error }, throwError(new Error()))
-    return drain(s).then(fail, e => assert.same(error, e))
+    return drain(s).then(rejects, eq(error))
```
briancavalier: See above.  Similarly with rejects vs. fail here.

packages/core/test/combinator/loop-test.js
```diff
@@ -24,8 +24,8 @@ describe('loop', function () {
 
     return collectEventsFor(a.length, s)
       .then(events => {
-        assert.same(a.length, events.length)
-        assert.equals([
+        is(a.length, events.length)
```
briancavalier: is() works fine here, but eq() is probably the right fit conceptually.  In general, I tend to think of eq() as the right thing for primitive values.

I feel like a good rule of thumb is:  Use eq() in all cases unless your intent is to test that two "pointers" point to the same object.

packages/core/test/combinator/loop-test.js
```diff
@@ -38,11 +38,11 @@ describe('loop', function () {
     const error = new Error()
     const s = loop(toPair, other, throwError(error))
 
-    return collectEventsFor(1, s).catch(e => assert.same(error, e))
+    return collectEventsFor(1, s).catch(eq(error))
```
briancavalier: See above about eq() vs is() for errors.

packages/core/test/combinator/merge-test.js
```diff
@@ -23,7 +22,7 @@ describe('merge', function () {
 
     return collectEventsFor(1, m1)
       .then(events1 => collectEventsFor(1, m2)
-        .then(events2 => assert.equals(events1, events2)))
+        .then(events2 => eq(events1, events2)))
```
briancavalier: `.then(eq(events1))` should work here.

packages/core/test/combinator/mergeConcurrently-test.js
```diff
@@ -21,13 +20,11 @@ describe('mergeConcurrently', () => {
     const n = 3
 
     return collectEvents(take(n, s), ticks(n))
-      .then(events => {
-        assert.equals(events, [
+      .then(eq([
```
briancavalier: Love the use of eq() here.  However, I think the objects below are indented an extra level.  I wonder why eslint isn't complaining ... hmmm.

packages/core/test/combinator/mergeConcurrently-test.js
```diff
@@ -36,8 +33,7 @@ describe('mergeConcurrently', () => {
     const n = 3
 
     return collectEvents(take(n * streams.length, s), ticks(n))
-      .then(events => {
-        assert.equals(events, [
+      .then(eq([
```
briancavalier: Same here, :+1: for eq(), but extra indentation below.

packages/core/test/combinator/mergeConcurrently-test.js
```diff
@@ -59,8 +54,7 @@ describe('mergeConcurrently', () => {
     const s = mergeConcurrently(m, fromArray(streams))
 
     return collectEvents(take(n * streams.length, s), ticks(m * n))
-      .then(events => {
-        assert.equals(events, [
+      .then(eq([
```
briancavalier: Same

packages/core/test/combinator/mergeConcurrently-test.js
```diff
@@ -70,15 +64,14 @@ describe('mergeConcurrently', () => {
...
+      ]))
   })
 })
 
 describe('mergeMapConcurrently', () => {
   it('when mapping function throws, it should catch and propagate error', () => {
     const error = new Error()
     const s = mergeMapConcurrently(x => { throw error }, 1, just(0))
-    return drain(s).then(fail, e => assert.same(error, e))
+    return drain(s).then(rejects, eq(error))
```
briancavalier: See above about rejects vs fail.

packages/core/test/combinator/promises-test.js
```diff
@@ -27,10 +26,10 @@ describe('promises', () => {
 
       return collectEventsFor(10, s)
         .then(events => {
-          assert.same(1, events.length)
+          is(1, events.length)
```
briancavalier: eq

packages/core/test/helper/stream-helper.js
```diff
@@ -0,0 +1,23 @@
+import { eq } from '@briancavalier/assert'
+import { reduce } from '../../src/combinator/reduce'
+
+export function assertSame (s1, s2) {
+  return Promise.all([toArray(s1), toArray(s2)]).then(arrayEquals)
+}
+
+export function expectArray (array, s) {
+  return toArray(s).then(function (result) {
```
briancavalier: oops, I think I missed this opportunity for partial application!  Can you update this to:

```js
return toArray(s).then(eq(result))
```

Thanks!

davidchase:
thanks for the comments, working on them now 😄

packages/core/test/combinator/concatMap-test.js
```diff
@@ -12,25 +12,23 @@ import { just as streamOf, never } from '../../src/source/core'
...
-}
+const identity = x => x
 
 describe('concatMap', function () {
   it('should satisfy associativity', function () {
     // m.concatMap(f).concatMap(g) ~= m.concatMap(function(x) { return f(x).concatMap(g); })
-    function f (x) { return streamOf(x + 'f') }
-    function g (x) { return streamOf(x + 'g') }
+    const f = x => streamOf(x + 'f')
+    const g = x => streamOf(x + 'g')
```
briancavalier: Oops, sorry, I missed this `streamOf` -> `just` opportunity.

packages/core/test/combinator/concatMap-test.js
```diff
@@ -96,18 +92,16 @@ describe('concatMap', function () {
 
   it('should dispose all inner streams', function () {
     const values = [1, 2, 3]
-    const spies = values.map(function () {
-      return spy()
-    })
+    const spies = values.map(() => spy())
 
-    const inners = values.map(function (x, i) {
+    const inners = values.map((x, i) => {
```
briancavalier: Can be an arrow expression, i.e. braces and `return` can be removed.

packages/core/test/combinator/sample-test.js
```diff
@@ -30,15 +30,15 @@ describe('sample', () => {
     const s = sample(Array, makeEvents(1, n), never())
 
     return collectEvents(s, ticks(n))
-      .then(events => is(0, events.length))
+      .then(events => eq(0, events.length))
```
briancavalier: OT: I've been thinking about adding map and contramap operations for assertions in `@briancavalier/assert`.  Here's a spot where it could be useful: we have an array, but we really want the the length.  A contramap op could transform the input.  Since assertions are just functions `a -> a`, contramap is just function composition "on the left":

```js
const length = array => array.length
// ...
return collectEvents(s, ticks(n))
  .then(pipe(length, eq(0)))
```

Anyway ... just thinking out loud :)

packages/core/test/helper/stream-helper.js
```diff
@@ -6,9 +6,7 @@ export function assertSame (s1, s2) {
 }
 
 export function expectArray (array, s) {
-  return toArray(s).then(function (result) {
-    eq(result, array)
-  })
+  return toArray(s).then(eq(array))
```
briancavalier: Nice catch!

packages/core/test/combinator/concatMap-test.js
```diff
@@ -96,18 +92,16 @@ describe('concatMap', function () {
...
       return new Stream(new FakeDisposeSource(spies[i], streamOf(x).source))
     })
 
     const s = concatMap.concatMap(identity, fromArray(inners))
 
-    return drain(s).then(function () {
-      spies.forEach(function (spy) {
+    return drain(s).then(() => {
+      spies.forEach(spy => {
```
briancavalier: These could both be arrow expressions, but honestly, I think it's more readable as is.

packages/core/test/combinator/concatMap-test.js
davidchase: woops thats on me

packages/core/test/combinator/concatMap-test.js
davidchase: yah wasnt sure if you had some standards on what you wanted to be an arrow expression based on length of definition or if its a one-liner, etc

packages/core/test/combinator/sample-test.js
davidchase: yes, yes that would be interesting... i love pushing the assert libraries and such to FP land 😍

packages/core/test/combinator/concatMap-test.js
davidchase: hmm yeah thats tough, i can see it as a 1-liner.. but i guess its to the 👁  of the beholder haha in this case you can make the call 😄

PR #43, opened 2017-02-17
Port top level tests

briancavalier:
Port all the tests in the top-level test dir.  Some of these should be moved into combinator, etc.  I think we should do that in another PR, since this one is already pretty big.

I also removed buster, and the entire test-buster-old dir.


TylorS:
:ship:

PR #46, opened 2017-03-05
Add newStream

briancavalier:
This adds a trivial constructor to make a stream from a `Sink -> Scheduler -> Disposable` function, which is all a stream is anyway now that #45 has been merged.  While someone *can* just make an object with a run() *method*, this feels a reasonable API to help folks.  It doesn't add any additional footgunnery beyond object-with-run-method, afaict, though it may encourage folks to try to write more imperative push streams ... maybe that's solvable via documentation?

### Todo

- [x] Add unit tests if this seems worthwhile.


TylorS:
:shipit:

briancavalier:
Added tests, and moved reduce, observe, and drain to the test/helpers folder, so they're no longer even a part of the main codebase 🎉 

@TylorS @davidchase can you guys take another quick look?  Thanks!

PR #49, opened 2017-04-02
chore(type-definitions): until

nissoh:


briancavalier:
Thanks @nissoh! I notice that skipUntil should also be named since. Would you mind including that change in this PR, too?

nissoh:
also noticed `until` in curried form will only accept `endSignal` type. @TylorS ?

packages/core/type-definitions/combinator/timeslice.d.ts
```diff
@@ -1,10 +1,10 @@
 import { Stream } from '@most/types';
 
 export function until<A>(signal: Stream<any>, s: Stream<A>): Stream<A>;
-export function until<A>(signal: Stream<A>): (s: Stream<A>) => Stream<A>;
+export function until<A>(signal: Stream<any>): (s: Stream<A>) => Stream<A>;
```
nissoh: also noticed until in curried form will only accept endSignal type. @TylorS ?

briancavalier:
Yep, `any` seems correct there.  Good catch.  This LGTM.  Any concerns, @TylorS?

packages/core/type-definitions/combinator/timeslice.d.ts
TylorS: It takes a `signal` then returns a function that takes a `stream` returning a new stream that will end when the signal emits.

PR #57, opened 2017-05-05
feat(withArrayValues): Add withArrayValues and zipArrayValues

briancavalier:
### Summary

This adds `withArrayValues` and `zipArrayValues` for mapping values from an array onto a stream.  In the case of `withArrayValues`, for each event, the stream provides the time component, and the array provides the value component.  Similarly for `zipArrayValues`, except it accepts a function to compute the value component from the array value and incoming event value.

### Other changes

- Run flow on the whole test dir.  It will only check files with `@flow` at the top, which for now, is only the new `withArrayValues-test.js`.  But, I like the idea of using unit tests also to check types.  Feedback welcome 😄 
  - Note that to do this, I had to give flow visibility into the top-level `node_modules` folder, since lerna is hoisting common dependencies (such as `mocha` and `@briancavalier/assert`) to the top level.

### Todo:

- [ ] deprecate `fromArray` and `fromIterable`

### DIFF
--- a/packages/core/src/combinator/withArrayValues.js
+++ b/packages/core/src/combinator/withArrayValues.js
@@ -6,7 +6,7 @@ export const withArrayValues = (array, stream) =>
   zipArrayValues(keepLeft, array, stream)
 
 export const zipArrayValues = (f, array, stream) =>
-  array.length === 0
+  array.length === 0 || stream === empty()
     ? empty()
     : new ZipArrayValues(f, array, stream)
 
--- a/packages/core/src/combinator/withArrayValues.js
+++ b/packages/core/src/combinator/withArrayValues.js
@@ -1,0 +1,44 @@
+/** @license MIT License (c) copyright 2010-2017 original author or authors */
+import { empty } from '../source/core'
+import Pipe from '../sink/Pipe'
+
+export const withArrayValues = (array, stream) =>
+  zipArrayValues(keepLeft, array, stream)
+
+export const zipArrayValues = (f, array, stream) =>
+  array.length === 0
+    ? empty()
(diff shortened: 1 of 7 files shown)


packages/core/src/combinator/withArrayValues.js
```diff
@@ -0,0 +1,44 @@
+/** @license MIT License (c) copyright 2010-2017 original author or authors */
+import { empty } from '../source/core'
+import Pipe from '../sink/Pipe'
+
+export const withArrayValues = (array, stream) =>
+  zipArrayValues(keepLeft, array, stream)
+
+export const zipArrayValues = (f, array, stream) =>
+  array.length === 0
```
briancavalier: Could probably also check if `stream === empty()` here.

briancavalier: Done in 27677af

packages/core/test/combinator/withArrayValues-test.js
```diff
@@ -0,0 +1,74 @@
+// @flow
+import { describe, it } from 'mocha'
+import { eq, fail } from '@briancavalier/assert'
+
+import { zipArrayValues, withArrayValues } from '../../src/combinator/withArrayValues'
+import { empty } from '../../src/source/core'
+
+import { collectEventsFor, makeEvents } from '../helper/testEnv'
+
+describe.only('withArrayValues', () => {
```
TylorS: there's a `.only` left here!

briancavalier: Removed in 772a7e3

davidchase:
👍

TylorS:
I really like having the tests type checked :+1:

PR #60, opened 2017-05-14
fix(types): missing withArrayValues

nissoh:


briancavalier:
0.6.1 published with this fix.

PR #63, opened 2017-05-31
feat(clock): Add relative monotonic clocks, platform detection

### LINKED ISSUE #62: Use monotonic relative time
Use relative time that starts at 0 instead of using system clock time.  Also, since the system clock isn't guaranteed to be monotonic, use performance.now or process.hrtime where available.

See: cujojs/most#452

briancavalier:
Close #62

Add relative monotonic clock implementations for performance.now and process.hrtime, and fallback clock for Date.now.  All clocks start at zero when created.

### Todo

- [x] Unit tests
- ~~Docs~~ will do in another PR

### DIFF
--- a/packages/scheduler/src/clock.js
+++ b/packages/scheduler/src/clock.js
@@ -2,7 +2,7 @@
 
 /*global performance, process*/
 
-class MillisecondClock {
+export class MillisecondClock {
   constructor (now, origin) {
     this.origin = origin
     this._now = now
@@ -13,7 +13,7 @@ class MillisecondClock {
   }
 }
 
-class HRTimeClock {
+export class HRTimeClock {
   constructor (hrtime, origin) {
     this.origin = origin
     this._hrtime = hrtime
--- a/packages/scheduler/src/clock.js
+++ b/packages/scheduler/src/clock.js
@@ -1,0 +1,48 @@
+/** @license MIT License (c) copyright 2010-2017 original author or authors */
+
+/*global performance, process*/
+
+export class MillisecondClock {
+  constructor (now, origin) {
+    this.origin = origin
+    this._now = now
+  }
+
+  now () {
+    return this._now() - this.origin
+  }
+}
+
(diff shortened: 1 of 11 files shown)


packages/scheduler/type-definitions/index.d.ts
```diff
@@ -1,9 +1,21 @@
...
 
 export function newScheduler (timer: Timer, timeline: Timeline): Scheduler;
 export function newScheduler (timer: Timer): (timeline: Timeline) => Scheduler;
 
 export function newDefaultScheduler (): Scheduler;
+
+export function newClockTimer (clock: Clock): Timer;
+export function newTimeline (): Timeline;
+
+declare function newPlatformClock (): Clock;
```
TylorS: each `declare` should be changed to `export`

briancavalier: Lol! I copied from flow and then proceeded to delete the wrong keyword from each line 🤣

briancavalier: Fixed!

TylorS:
I think this is good to go 👍

briancavalier:
Cool.  Gonna write tests for the clock module and then I think this will be ready to merge.

briancavalier:
At some point we need to tackle some minimal docs for `@most/scheduler`.  I'll try to do that in another PR.

packages/scheduler/src/clock.js
```diff
@@ -0,0 +1,48 @@
...
+    const hrt = this._hrtime(this.origin)
+    return (hrt[0] * 1e9 + hrt[1]) / 1e6
+  }
+}
+
+export const millisecondClockFromNow = now =>
+  new MillisecondClock(now, now())
+
+export const newPerformanceNowClock = () =>
+  millisecondClockFromNow(performance.now)
```
frangio: This should be `performance.now.bind(performance)`. It's failing for me on Firefox.

PR #64, opened 2017-05-31
feat(prelude): move prelude into monorepo

TylorS:
This is largely just a copy and paste from the current repo.


TylorS:
closing in favor of #65

PR #65, opened 2017-05-31
feat(prelude): imports prelude into monorepo

TylorS:
Using `lerna import` pulls `@most/prelude` into the monorepo.
Minor changes include:
-removing `jsinspect` since it no longer supports node 4
-removing babel now that we are using buble across the board


packages/prelude/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@most/prelude",
+  "version": "1.5.2",
```
briancavalier: Hmmm, I didn't think about versioning implications.  Anyone know the right way to handle this in lerna, given that we're not currently using `--independent`?

TylorS: Hmmm, I actually don't know. Anything wrong with switching to independent mode?

briancavalier: I don't think so ... seems simpler not to, but if we have to, that's probably ok.

briancavalier:
Looks good.  Should we do this before 1.0?  Also, I guess we'll need to update the [mostjs/prelude](https://github.com/mostjs/prelude) repo with a message saying it has moved here.

PR #66, opened 2017-06-03
feat(core): POC Use stream-local relative time

briancavalier:
This PR switches to stream-local time.  Each stream effectively has its own virtual clock that starts at zero.  In reality, they all share the same clock, but have an associated origin time offset.  This is done via a simple RelativeScheduler that wraps Scheduler and adds an offset.  Higher-order streams create a new RelativeScheduler, based on their own scheduler, with the appropriate time offset, and use that to run each inner stream.

Time must also be transformed "in the other direction", from local time to outer time, when merging inner events back to the outer sink.  RelativeSink implements that transform.  It was also convenient to implement it directly in the custom sinks used by mergeConcurrently and switchLatest.  It might be better to take the hit on creating a new extra RelativeSink wrappers rather than implement the transform in multiple places.

An analogy is 2d and 3d graphics, where it is desirable to be able to program objects, sprites, characters, etc. each of which has its own coordinate system, and there are transforms to and from local coord systems to scene/world coords (and finally to screen coords).

### Todo

- [ ] Types
- [ ] Unit tests
- [ ] Docs

### DIFF
--- a/packages/core/src/combinator/continueWith.js
+++ b/packages/core/src/combinator/continueWith.js
@@ -3,6 +3,7 @@
 /** @author John Hann */
 
 import Pipe from '../sink/Pipe'
+import RelativeSink from '../sink/RelativeSink'
 import { disposeOnce, tryDispose } from '@most/disposable'
 
 export const continueWith = (f, stream) =>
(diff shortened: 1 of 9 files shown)


briancavalier:
Closing in favor of #67

Issue #10, opened 2017-01-05
Add curryable zip2/zip3/etc, combine2/combine3/etc?


briancavalier:
One way to solve the "problem" of variadic functions not being curry-friendly is to provide variants with specific arities.  Haskell does this, i.e. zip2, zip3, et al.  We could do the same by keeping the variadic versions and simply providing fixed-arity wrappers.  For example:

```js
import { combine } from './combinators/combine'

export { combine } // variadic

export combine2 = curry3((f, s1, s2) => combine(f, s1, s2))
export combine3 = // ...
// etc
```

Of course, there's always the option of waiting until folks ask for this before actually implementing it.  But, I wanted to get some current thoughts on it.

TylorS:
I think it'd be easier to have `combine(f, s1, s2)` as a non-variadic function, but then to simply have `combineArray(f, [ ...streams ])`. A more radical approach would to remove `combine` altogether in favor of `combineArray`, which then, might be able to simple be `combine(f, [ ...streams])`.

davidchase:
I think i kinda like the `combine(f, s1, s2)` and also the `combineArray(f, [... streams])`, though i am guessing knowing the arity (`zip2, zip3`) ahead of time is better for the program? so something like `combineN(2, f, s1, s2)` would not be ideal ?

jethrolarson:
I don't mind `combine2` etc. 
With ramda I think I'd rather have curry3 than curryN(3), though that means they all have to be documented and tested which may have diminishing returns. 

BTW is combine2 basically the same as liftA2?

briancavalier:
> is combine2 basically the same as liftA2?

Yep.  Variadic `combine` is basically variadic `liftA`

> they all have to be documented and tested which may have diminishing returns

Good point.  I think you're right about diminishing returns ... combine4 is probably more rare in the wild than combine3, and so on.  Going with 2 and 3 is probably enough for the vast majority cases.  Since they'd all be implemented using variadic `combine`, the additional tests could be pretty minimal.  I'll bet docs could be similarly minimal: "A fixed argument version of combine that allows partial application" or somesuch might be close to enough.

The thing that variadic `combine/zip` *and* array-based `combineArray/zipArray` each lack is partial application, i.e. I can't fix the function and the first stream and demand the second, or fix only the function to lift a 2-argument function *which is also partially applicable*. without using a partial application helper (or `bind`).  Partial application aside, I don't think there's any benefit to providing fixed-arity versions.

[...]

PR #68, opened 2017-06-04
Rewrite combine, merge, & zip to be curryable

joshburgess:
I made a PR to address https://github.com/mostjs/core/issues/10

I changed combine, curry, and zip to only take in two stream arguments, rather than a variable amount, in order to make them curryable. Please, let me know if anything needs to be changed.


joshburgess:
There seems to be some problem with ESLint detecting an error here, saying that `_merge` is undefined for some reason, but I don't see why... Any ideas?

```
// -----------------------------------------------------------------------
// Merging

export { merge as _merge, mergeArray } from './combinator/merge'

export const merge = curry2(_merge)
```

TylorS:
```
import { merge as _merge, mergeArray } from './combinator/merge'

export const merge = curry3(_merge)
export { mergeArray }

```
This would be correct there, switch the `export` keyword with `import`

joshburgess:
@TylorS Woops. It was 4:00 AM when I pushed up those commits changes. Sorry. 😂 Fixing now...

briancavalier:
Thanks @joshburgess.  I only gave this a quick glance, and the code changes look good.  I'll take a closer look at the code later today.  I noticed two other things, tho:

1. The TS and Flow type defs for `merge` should only allow merging homogenous stream types.  Heterogenous merging gets out of hand quickly.  So, `merge` should only allow merging `Stream<A>` with `Stream<A>` to produce another `Stream<A>`.
2. A yarn.lock file snuck into the PR.  We're probably going to switch to yarn at some point, but we can't until lerna and yarn work together seamlessly.

joshburgess:
@briancavalier Thanks for the feedback. I've fixed those issues now. I didn't even realize the yarn file had snuck in there. I must have used yarn to install the npm packages out of habit and not noticed it. I should probably stop putting together PRs at 4:00 in the morning. lol. Let me know if anything else needs changing.

packages/core/src/combinator/zip.js
```diff
@@ -7,20 +7,20 @@ import { empty } from '../source/core'
...
-import { map as mapArray, tail } from '@most/prelude'
+import { map as mapArray } from '@most/prelude'
 import invoke from '../invoke'
 import Queue from '../Queue'
 
 /**
- * Combine streams pairwise (or tuple-wise) by index by applying f to values
- * at corresponding indices.  The returned stream ends when any of the input
- * streams ends.
+ * Combine two streams pairwise (or tuple-wise) by index by applying f to
```
briancavalier: We can remove "(or tuple-wise)" here since it's always exactly two now.

packages/core/src/combinator/zip.js
```diff
@@ -7,20 +7,20 @@ import { empty } from '../source/core'
...
+ * values at corresponding indices.  The returned stream ends when any of
```
briancavalier: We can be more specific here now: "The returned stream ends with either of" (any -> either)

packages/core/src/index.js.flow
```diff
@@ -74,7 +74,7 @@ declare export function continueWith <A> (f: (any) => A): (s: Stream<A>) => Stre
 
 declare export function switchLatest <A> (s: Stream<Stream<A>>): Stream<A>
 
-declare export function merge <A> (...ss: Array<Stream<A>>): Stream<A>
+declare export function merge <A> (sa: Stream<A>, sb: Stream<A>): Stream<A>
```
briancavalier: Types look good :+1:.  In a few places, we've already used `sb` as a name for `Stream<B>`.  How about we call these two streams `s1` and `s2`, or `sa1` and `sa2`, just to avoid any potential confusion.

Also need to add the other curried type def variant here to allow partial application:

```js
declare export function merge <A> (sa: Stream<A>): (sb: Stream<A>) => Stream<A>
```

For two parameters, it's not too bad ... but for 3 it starts to get tedious 😞 See below.

packages/core/src/index.js.flow
```diff
@@ -83,10 +83,6 @@ declare export function sample <A, B, C> (f: (A, B) => C, sampler: Stream<A>): (
 declare export function sample <A, B, C> (f: (A, B) => C): (sampler: Stream<A>) => (sampled: Stream<B>) => Stream<C>
 
 declare export function combine <A, B, Z> (f: (A, B) => Z, sa: Stream<A>, sb: Stream<B>): Stream<Z>
```
briancavalier: Need to add the other partial application variants for combine and zip.  Take a look at [the scan type defs for an example](https://github.com/mostjs/core/blob/master/packages/core/src/index.js.flow#L43-L46).

packages/core/type-definitions/combinator/combine.d.ts
```diff
@@ -1,10 +1,6 @@
 import { Stream } from '@most/types';
 
 export function combine<A, B, R>(fn: (a: A, b: B) => R, a: Stream<A>, b: Stream<B>): Stream<R>;
```
briancavalier: Similarly for the TS type defs, need to add the partial application variants for `merge`, `combine`, and `zip`.

packages/core/type-definitions/combinator/merge.d.ts
```diff
@@ -1,4 +1,4 @@
 import { Stream } from '@most/types';
 
-export function merge<A>(...streams: Array<Stream<A>>): Stream<A>;
+export function merge<A>(a: Stream<A>, b: Stream<A>): Stream<A>;
```
briancavalier: Whatever names we pick for these two streams in the `merge` flow type refs, let's use the same names here.

packages/core/src/combinator/merge.js
```diff
@@ -6,15 +6,14 @@ import Pipe from '../sink/Pipe'
...
- * @returns {Stream} stream containing events from all streams in the argument
- * list in time order.  If two events are simultaneous they will be merged in
- * arbitrary order.
+ * @returns {Stream} stream containing events from two streams in time order.
+ * If two events are simultaneous they will be merged in arbitrary order.
  */
-export function merge (/* ...streams*/) {
-  return mergeArray(copy(arguments))
+export function merge (stream1, stream2) {
+  return mergeArray([stream1, stream2])
```
briancavalier: And here :+1:

packages/core/src/combinator/zip.js
```diff
@@ -7,20 +7,20 @@ import { empty } from '../source/core'
...
+ * the input streams ends.
  * @param {function} f function to combine values
  * @returns {Stream} new stream with items at corresponding indices combined
  *  using f
  */
-export function zip (f /*, ...streams */) {
-  return zipArray(f, tail(arguments))
+export function zip (f, stream1, stream2) {
+  return zipArray(f, [stream1, stream2])
```
briancavalier: And here! :+1:

packages/core/src/combinator/combine.js
```diff
@@ -7,17 +7,16 @@ import { empty } from '../source/core'
...
- * Combine latest events from all input streams
+ * Combine latest events from two streams
  * @param {function(...events):*} f function to combine most recent events
  * @returns {Stream} stream containing the result of applying f to the most recent
  *  event of each input stream, whenever a new event arrives on any stream.
  */
-export function combine (f /*, ...streams */) {
-  return combineArray(f, tail(arguments))
+export function combine (f, stream1, stream2) {
+  return combineArray(f, [stream1, stream2])
```
briancavalier: :heart: So nice to get rid of the use of `arguments`!

packages/core/src/combinator/merge.js
joshburgess: With these "And here 👍 " comments, were you saying they looked good or did you want me to rename stream1 & stream2? Sorry, just trying to follow along before I make the changes.

briancavalier: Ugh, sorry!  Github reordered my comments.  The "And here" comments were referring to my comment about being glad to get rid of `arguments`.

joshburgess:
Ok, @briancavalier , I believe I addressed everything now with the latest commits. Anything else?

PR #72, opened 2017-06-04
fix(clock): Call performance.now as a method. Add new RelativeClock

### LINKED ISSUE #69: TypeError calling performance.now
I mentioned it in [a comment](https://github.com/mostjs/core/pull/63/files#r120018903) in a closed PR so I'm opening an issue here.

The `performance.now` method is being called as a function without binding `this` to `performance` and it's causing an error. I guess this wasn't detected in testing because the tests run in a node.js environment?

briancavalier:
Fix #69 
Close #71 

Consider `Date` and `performance` to be acceptable `Clock` types.   Add new RelativeClock that accepts any `Clock` and creates a new `Clock` relative to its current time.

### DIFF
--- a/packages/scheduler/src/clock.js
+++ b/packages/scheduler/src/clock.js
@@ -28,10 +28,10 @@ export class HRTimeClock {
 export const relativeClock = clock =>
   new RelativeClock(clock, clock.now())
 
-export const newPerformanceNowClock = () =>
+export const newPerformanceClock = () =>
   relativeClock(performance)
 
-export const newDateNowClock = () =>
+export const newDateClock = () =>
   relativeClock(Date)
 
 export const newHRTimeClock = () =>
@@ -39,10 +39,10 @@ export const newHRTimeClock = () =>
 
 export const newPlatformClock = () => {
   if (typeof performance !== 'undefined' && typeof performance.now === 'function') {
-    return newPerformanceNowClock()
+    return newPerformanceClock()
   } else if (typeof process !== 'undefined' && typeof process.hrtime === 'function') {
     return newHRTimeClock()
   }
 
-  return newDateNowClock()
+  return newDateClock()
 }
--- a/packages/scheduler/src/clock.js
+++ b/packages/scheduler/src/clock.js
@@ -25,14 +25,14 @@ export class HRTimeClock {
   }
 }
 
-export const relativeClock = clock =>
+export const clockRelativeTo = clock =>
   new RelativeClock(clock, clock.now())
 
 export const newPerformanceClock = () =>
-  relativeClock(performance)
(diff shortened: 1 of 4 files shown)


packages/scheduler/src/clock.js
```diff
@@ -2,37 +2,37 @@
...
   now () {
-    const hrt = this._hrtime(this.origin)
+    const hrt = this.hrtime(this.origin)
     return (hrt[0] * 1e9 + hrt[1]) / 1e6
   }
 }
 
-export const millisecondClockFromNow = now =>
-  new MillisecondClock(now, now())
+export const relativeClock = clock =>
```
frangio: Should this be called `newRelativeClock` for consistency with the other exported functions?

packages/scheduler/src/clock.js
```diff
@@ -2,37 +2,37 @@
...
+  new RelativeClock(clock, clock.now())
 
 export const newPerformanceNowClock = () =>
```
frangio: How about `newPerformanceClock`?

packages/scheduler/src/clock.js
```diff
@@ -2,37 +2,37 @@
...
-  millisecondClockFromNow(performance.now)
+  relativeClock(performance)
 
 export const newDateNowClock = () =>
```
frangio: ... and `newDateClock`?

packages/scheduler/src/clock.js
briancavalier: We've been using `new*` for creating a new thing from nothing, or from its component pieces.  This case is kind of interesting because we're transforming a clock.  It's more like `Array#map` than `new Array`.  Any suggestions on a name that means "transform a clock"?

Ideas: `relativeTo`, `withOrigin`

packages/scheduler/src/clock.js
briancavalier: Good call :+1:

packages/scheduler/src/clock.js
frangio: Oh, I see. It's actually more like `Array#from` IMO. So with that as an idea, how about `relativeClockFrom`? I'm not sure that it fits with most's naming conventoins, or that it's better than just `relativeClock`.

briancavalier: Good suggestion, thanks.  It spawned quite a bit of discussion among the core team!  We finally decided on `clockRelativeTo`.  I think we'll probably add a similar `schedulerRelativeTo` in #67.

briancavalier:
@frangio We just published `@most/scheduler` and `@most/core` 0.7.1 with this fix.  When you have a chance, please let us know if it's working for you.

PR #74, opened 2017-06-07
chore(switchLatest): avoid mutating Segment

nissoh:


briancavalier:
Thanks @nissoh.  Taking a look now.

packages/core/src/combinator/switch.js
```diff
@@ -82,12 +81,12 @@ class SwitchSink {
...
 
 class Segment {
-  constructor (min, max, outer, sink) {
+  constructor (source, min, max, outer, sink, scheduler) {
     this.min = min
     this.max = max
     this.outer = outer
     this.sink = sink
-    this.disposable = disposeNone()
+    this.disposable = source.run(sink, scheduler)
```
briancavalier: Should pass `this` here instead of `sink`.  Looking at the original code, `this.current` is passed to `stream.run`, and `this.current` is the Segment.  So, in the new code, the Segment needs to pass itself to `source.run`.

Interesting that it didn't cause a unit test failure ... seems like a hole in our tests.

PR #75, opened 2017-06-13
chore(unfold): Remove unfold, iterate, generate

briancavalier:
Remove unfold, iterate, and generate.  These are infrequently-used, are a bit loose with their types, since they allow user-provided functions to return a promise or not, and thus are not very rigorous about time.

If there's a demand for this, I can see reintroducing them as a separate package that requires returning promises, for example.


TylorS:
:+1+

PR #76, opened 2017-06-14
fix(newStream): export newStream from index.js

axefrog:
newStream declared in TypeScript defs but not matched by an actual export.


briancavalier:
Thanks!

PR #77, opened 2017-06-16
feat(multicast): adds multicast back to the core

davidchase:
It seems like we need to figure out if we are going with this direction (as this PR has it) and keeping `@most/multicast` published and on npm
or are we all leaning to just bringing it back and storing it with the combinators.


axefrog:
Worth noting that the current version of multicast still targets the old version of Most. I imagine that would need to be fixed prior to this PR being accepted.

axefrog:
Heh looks like I saw this earlier and forgot about it :D

packages/core/src/combinator/multicast.js
```diff
@@ -0,0 +1,98 @@
+import { append, remove, findIndex } from '@most/prelude'
+import { tryEvent, tryEnd } from '../source/tryEvent'
+
+const dispose = disposable => disposable.dispose()
+
+const emptyDisposable = {
+  dispose () {}
+}
```
briancavalier: We can use [`disposeNone` from `@most/disposable`](https://github.com/mostjs/core/blob/master/packages/disposable/src/disposeNone.js).

packages/core/src/combinator/multicast.js
```diff
@@ -0,0 +1,98 @@
...
+    this.disposed = false
+  }
+
+  dispose () {
+    if (this.disposed) {
+      return
+    }
+    this.disposed = true
+    const remaining = this.source.remove(this.sink)
+    return remaining === 0 && this.source._dispose()
```
briancavalier: We should probably update `dispose` to return undefined that it's strict with types, e.g.

```js
    if(this.source.remove(this.sink) === 0) {
      this.source._dispose()
    }
```

packages/core/src/combinator/multicast.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
```
briancavalier: MulticastDisposable includes some of the functionality of [`disposeOnce`](https://github.com/mostjs/core/blob/master/packages/disposable/src/disposeOnce.js).  We could potentially refactor it to just deal with multicast dispose concerns and then wrap it with `disposeOnce`.  E.g. it could be something like:

```js
class MulticastDisposable {
  constructor (source, sink) {
    this.source = source
    this.sink = sink
  }

  dispose () {
    if(this.source.remove(this.sink) === 0) {
      this.source._dispose()
    }
  }
}
```

And then below:

```js
return disposeOnce(new MulticastDisposable(this, sink))
```

I don't think it's strictly necessary for this PR, but could be a nice simplification.  What do you think?

packages/core/src/combinator/multicast.js
```diff
@@ -0,0 +1,98 @@
...
+    if (n === 1) {
+      this._disposable = this.source.run(this, scheduler)
+    }
+    return new MulticastDisposable(this, sink)
+  }
+
+  _dispose () {
+    const disposable = this._disposable
+    this._disposable = emptyDisposable
+    return Promise.resolve(disposable).then(dispose)
```
briancavalier: Don't need to deal with promises here anymore, so this can be simplified:

```js
return disposable.dispose()
```

packages/core/src/combinator/multicast.js
```diff
@@ -0,0 +1,98 @@
...
+  constructor (source) {
+    this.source = new MulticastSource(source)
+  }
+  run (sink, scheduler) {
+    this.source.run(sink, scheduler)
+  }
+}
+
+export const multicast = stream =>
+  stream instanceof Multicast ? stream : new Multicast(stream)
```
briancavalier: Let's reorder this file to follow the order used by other combinators: exports at the top, followed by supporting classes and functions.  E.g. `multicast` export at the top, followed by `Multicast`, followed by `MulticastSource`.

packages/core/test/combinator/multicast-test.js
```diff
@@ -0,0 +1,37 @@
...
+    eq(s, multicast(s))
+  })
+
+  it('should call mapper function once when there are > 1 observer', () => {
+    const mapperSpy = spy()
+    const observer1Spy = spy()
+    const observer2Spy = spy()
+    const observer3Spy = spy()
+
+    const s = now({value: 'sentinel'})
```
briancavalier: When I saw `{ value: 'sentinel' }`, I went looking for an assertion about the the value.  Since the event value is never used, it could something like `1` or `{}`.

packages/core/test/combinator/multicast-test.js
```diff
@@ -0,0 +1,37 @@
...
+    const mapped = map(mapperSpy, s)
+    const multicasted = multicast(mapped)
+
+    const o1 = observe(observer1Spy, multicasted)
+    const o2 = observe(observer2Spy, multicasted)
+    const o3 = observe(observer3Spy, multicasted)
+
+    Promise.all([o1, o2, o3]).then(() => {
```
briancavalier: Missing `return` ... d'oh, it's [missing here too](https://github.com/mostjs/multicast/blob/master/test/index-test.js#L27)!

packages/core/test/combinator/multicast-test.js
```diff
@@ -0,0 +1,37 @@
...
+      assert(mapperSpy.calledOnce)
+      eq(true, observer1Spy.calledOnce)
+      eq(true, observer2Spy.calledOnce)
+      eq(true, observer3Spy.calledOnce)
```
briancavalier: Hmmm, ok, looking more closely at these assertions, we may want to strengthen them a bit.  What if we did this:

```js
const f = x => x + 1
const x = Math.random()
const s = now(x)

const mapperSpy = spy(f)
const mapped = map(mapperSpy, s)

// ...

return Promise.all([o1, o2, o3]).then(() => {
  assert(mapperSpy.calledOnce)

  const expected = f(x)

  assert(observer1Spy.calledOnce)
  eq(expected, observer1Spy.getCall(0).args[0])

  assert(observer2Spy.calledOnce)
  eq(expected, observer2Spy.getCall(0).args[0])

  assert(observer2Spy.calledOnce)
  eq(expected, observer2Spy.getCall(0).args[0])
})
```

packages/core/test/combinator/multicast-test.js
```diff
@@ -0,0 +1,37 @@
...
+    const observer2Spy = spy()
+    const observer3Spy = spy()
+
+    const s = now({value: 'sentinel'})
+    const mapped = map(mapperSpy, s)
+    const multicasted = multicast(mapped)
+
+    const o1 = observe(observer1Spy, multicasted)
+    const o2 = observe(observer2Spy, multicasted)
+    const o3 = observe(observer3Spy, multicasted)
```
briancavalier: It'd be nice to use `collectEvents` or `runEffects` instead of `observe`.  Maybe something like:

```js
// import { runEffects } from '../../src/runEffects'
// import { ticks } from '../helper/testEnv'
const scheduler = ticks(1)

const o1 = runEffects(tap(observerSpy1, multicasted), scheduler)
const o2 = runEffects(tap(observerSpy2, multicasted), scheduler)
const o3 = runEffects(tap(observerSpy3, multicasted), scheduler)
```

I didn't try it, but I think that's close.

packages/core/src/combinator/multicast.js
Frikki: INHO, a source file should read like a book, i.e., top-down, which also follows the clean code conventions. E.g.:

```
const emptyDisposable = {
  dispose()
}

const dispose = disposable => disposable.dispose()
```

... such that `dispose` is follows after its first usage. This prevents readers of the file to unnecessarily scroll up and down in a search for declarations but instead, readers can scroll only down and allow for a fluent experience similar to any other form of reading.

briancavalier: Agreed. I think we're saying the same thing, but want to make sure. You've said it much more eloquently, and highlighted the "why".

davidchase: For me top down has meant first you have the definitions and then followed by usage. So usually in some of my codebases you scroll to the bottom to get the exports and the top of the file is the source for the exports. I know thats a bit different here.

packages/core/src/combinator/multicast.js
```diff
@@ -0,0 +1,89 @@
...
+
+class Multicast {
+  constructor (source) {
+    this.source = new MulticastSource(source)
+  }
+  run (sink, scheduler) {
+    this.source.run(sink, scheduler)
```
briancavalier: Need a `return` to avoid breaking the disposable chain:

```js
return this.source.run(sink, scheduler)
```

packages/core/test/combinator/multicast-test.js
```diff
@@ -0,0 +1,49 @@
...
+import { assert, eq } from '@briancavalier/assert'
+import { spy } from 'sinon'
+
+import { now } from '../../src/source/now'
+import { map, tap } from '../../src/combinator/transform'
+import { multicast } from '../../src/combinator/multicast'
+import { runEffects } from '../../src/runEffects'
+import { ticks } from '../helper/testEnv'
+
+describe('multicast', () => {
```
briancavalier: Over in the mostjs/multicast repo, there are [tests for MulticastSource and MulticastDisposable](https://github.com/mostjs/multicast/tree/master/test).  I think it's important to port those as well.  You'll just need to export MulticastSource and MulticastDisposable for the purposes of unit testing them.  I'd be fine with copying those tests into `multicast-test.js`.

It kinda raises the question of whether we should split up multicast.js and multicast-test.js, but I'm ok with *not* splitting them up in this PR.

Also, on the topic of exports, currently [`@most/multicast` exports MulticastSource](https://github.com/mostjs/multicast/blob/master/src/index.js#L10).  It seems like we may also want to export it from `@most/core`.  Thoughts?

packages/core/src/combinator/multicast.js
davidchase: good catch 👀

packages/core/test/combinator/multicast-test.js
davidchase: Im guessing for something like most-subect we should export MulticastSource?

will port over the other tests into multicast-test.js for the time being

briancavalier: > Im guessing for something like most-subect we should export MulticastSource?

:+1:

packages/core/src/combinator/multicast.js
```diff
@@ -75,14 +75,14 @@ class MulticastSource {
...
-class MulticastDisposable {
+export class MulticastDisposable {
   constructor (source, sink) {
     this.source = source
     this.sink = sink
   }
 
   dispose () {
-    if (this.source.remove(this.sink === 0)) {
+    if (this.source.remove(this.sink) === 0) {
```
briancavalier: 😲

packages/core/test/combinator/multicast-test.js
```diff
@@ -0,0 +1,191 @@
...
+
+    const f = x => x + 1
+    const x = Math.random()
+    const s = now(x)
+
+    const mapperSpy = spy(f)
+    const mapped = map(mapperSpy, s)
+    const multicasted = multicast(mapped)
+
+    const o1 = runEffects(tap(observer1Spy, multicasted), ticks(1))
```
briancavalier: A lot of these tests use separate schedulers now (i.e. Calling ticks multiple times in the same test). The tests pass, but it's not representative of normal/supported use cases, where you'd have exactly 1 scheduler per application.

So, I think the right thing is for each test to create 1 scheduler, like 'should publish events to all observers' does below.

packages/core/src/combinator/multicast.js
davidchase: Yup that will do it

packages/core/test/combinator/multicast-test.js
davidchase: yeah that makes sense updating 👍

PR #78, opened 2017-06-19
chore(docs): Add readthedocs documentation scaffold

### LINKED ISSUE #22: API docs
I'd really love to try something new here, but I don't know if it's feasible.  Since we'll have a flow type definition file after #21 is merged, I wonder if we could find a way to write markdown comments in it and generate an API doc from that.  I'm thinking something very simple, along the lines of [docco](https://jashkenas.github.io/docco/), to generate markdown or HTML.

Thoughts?  Other ideas?

briancavalier:
Close #22 

This PR adds a basic scaffold for readthedocs.  It gives us a framework in which we can start adding docs.  The config is set to build the this branch (`add-rtd`), so we need to switch that before merging.

### Todo

- [ ] Switch config to build master branch (on readthedocs.org)
- [ ] Revise index.rst and setup basic TOC


briancavalier:
Readthedocs can support both reStructuredText and Markdown.  So, if we would rather write in markdown, we can.  Markdown files can't use the extra features of rst, tho, like TOC generation, includes, references, etc.

For example, we could use rst for the main index doc, which will have a TOC.  Then, we could use markdown for other files.

Or, we can just go all-in on rst--might be easier to pick one format and just go with it.

davidchase:
> Or, we can just go all-in on rst--might be easier to pick one format and just go with it.

I think just pick one `rst` and go with it if thats a possibility especially if it has more features, we just probably need a rst cheatsheet when writing docs/et al

PR #79, opened 2017-06-19
feat(at): Add at and now. Move empty, never to own modules

briancavalier:
Add `at(time, value)` and `now(value)`, per recent discussion around ensuring combinators always have meaningful time semantics and names.  With the recent move to event stream-local clocks, `at(t, x)` has semantics of "x occurs at time t on the newly created event stream's local clock", and `now(x)` has the same semantics, with t fixed to zero.

I didn't remove `just` (or the `src/source/core.js` module) yet, since it's used *a lot* in unit tests, and I didn't want to deal with the collateral damage until we know for sure this is the way we want to go.

Definitely looking for feedback and discussion!

### Other changes

Other changes we may want to keep regardless of now/at.

1. Moved `empty` and `never` to their own modules.
2. Updated some deps

### Todo

- [x] Remove `just`, and `src/source/core.js`
- [x] Add type defs for `at` and `now`


packages/core/src/source/at.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  run (sink, scheduler) {
+    return scheduler.delay(this.time, propagateTask(runAt, this.value, sink))
+  }
+}
+
+function runAt (t, x, sink) {
+  sink.event(t, x)
+  sink.end(t, undefined)
```
briancavalier: Ah, can remove the `undefined` here now since we dropped explicit `end` values.  Will do that later today.

Frikki: How could I even miss that? Usually, I have a keen eye for `null` and `undefined`. Values I personally avoid at all costs.

briancavalier:
🔪 `just` 🔪 

Removing it made the PR very noisy, but such is the price of progress 😃

briancavalier:
I think this is ready for another review.  Sorry for the noise around removing `just`.

PR #80, opened 2017-06-19
chore(concat): Remove concat

briancavalier:
Remove concat.  The name "concat" conjures up a mental model of array concatenation, which can have completely nonsensical time semantics.  For example, what does *concatenating* mouse clicks onto mousemoves even mean?  It also seems to have no practical use outside of *data processing* applications, which are not the focus of `@most/core`.

Interestingly, `continueWith` seems to have a more reasonable mental model: start by observing event stream 1 and when it ends, start observing event stream 2.  To use the mouse click/move example from above, asking for all the mouse clicks that happen after some number of mouse moves, e.g. `continueWith(() => click(el), take(10, mousemove(el)))`, at least seems to conjure up a sensible mental image.

It's possible a similar operator is still useful, but with clearer time semantics, and a name that creates a mental image like "all the Ys that happen after all the Xs", like `continueWith` seems to do.


Frikki:
I totally agree with that.

briancavalier:
Gonna merge this after #79, since that seems like it will cause less of a rebase headache.

PR #81, opened 2017-06-23
chore(coverage): Enable nyc coverage in all packages

briancavalier:
Coverage was enabled in `@most/prelude`, but not other packages.  This adds coverage reporting to the rest of the packages and updates nyc to the latest version.

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -31,7 +31,6 @@
   "author": "brian@hovercraftstudios.com",
   "license": "MIT",
   "devDependencies": {
-    "flow-bin": "^0.47.0",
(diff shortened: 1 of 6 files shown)


packages/types/package.json
```diff
@@ -31,6 +31,7 @@
   "author": "brian@hovercraftstudios.com",
   "license": "MIT",
   "devDependencies": {
-    "flow-bin": "^0.47.0"
+    "flow-bin": "^0.47.0",
+    "nyc": "^11.0.2"
```
briancavalier: Hmmm, don't need this.  I'll remove it.

PR #84, opened 2017-06-28
docs(combinator-template): Add combinator doc template

briancavalier:
This adds a new page, "Writing docs", where we could put helpful information for how to write docs.  Initially, it just includes a template that can be cut and pasted into api.rst when writing docs for an operator.


davidchase:
:sheep::it:

PR #85, opened 2017-07-03
feat(runWithLocalTime): avoid nesting RelativeSinks

briancavalier:
Accumulate RelativeSink offsets to avoid nesting.

I don't like `instanceof`, but in this case it feels like the right thing.  The alternative would be to require a new method (i.e. analogous to Scheduler's recently-added `relative` method) on all Sinks, which feels too annoying for people writing custom Sinks.  Alternatively, we could offer a superclass that provides the method, but I dislike coupling people's code to a superclass even more than I dislike `instanceof`.

### DIFF
--- a/packages/core/src/scheduler/runWithLocalTime.js
+++ b/packages/core/src/scheduler/runWithLocalTime.js
@@ -3,5 +3,18 @@
 import RelativeSink from '../sink/RelativeSink'
 import { schedulerRelativeTo } from '@most/scheduler'
 
+// Run a stream with its own localized clock
+// This transforms time from the provided scheduler's clock to a stream-local
+// clock (which starts at 0), and then *back* to the scheduler's clock before
+// propagating events to sink.  IOW, stream.run will see local times, and sink
+// will see scheduler times.
 export const runWithLocalTime = (origin, stream, sink, scheduler) =>
-  stream.run(new RelativeSink(origin, sink), schedulerRelativeTo(origin, scheduler))
+  stream.run(relativeSink(origin, sink), schedulerRelativeTo(origin, scheduler))
+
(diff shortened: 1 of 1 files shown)


Frikki:
However, doesn’t the `sink` parameter becomes ambiguous?

briancavalier:
> doesn’t the sink parameter become ambiguous?

I'm not sure what you mean.  Ambiguous in what way?  Could you explain a bit more?

Frikki:
The type of `sinks`.

briancavalier:
@Frikki Here's why I think the type of `sink` is unambiguous.

I'm viewing `Sink` as a typeclass/interface (e.g. it's [a TS `interface` in `@most/types`](https://github.com/mostjs/core/blob/master/packages/types/index.d.ts#L10)), of which `RelativeSink` is one particular implementation, and the type of `relativeSink` as:

```hs
relativeSink :: Sink s => s a -> RelativeSink a
```

IOW, given any `Sink` which propagates `a`s, `relativeSink` returns a `RelativeSink` that propagates `a`s.  So, at it's interface, `relativeSink` is well-typed (even though it's *implementation* does use `instanceof` to make a runtime type discrimination--see PR description for why I think this is the lesser of all evils).  The `sink` passed to `runWithLocalTime` and `stream.run` has the same type constraint: It can be any `Sink a`.

Does that make sense?

Frikki:
@briancavalier Indeed, it does. I was looking for the type definition before your last comment but didn’t find it. Everything is dandy and fine.

briancavalier:
Cool.  I think you highlighted one of the problems with our having externally facing type definitions, and not (yet) having them internally.  Types are great internal documentation.

PR #87, opened 2017-07-11
feat(types): Add explicit covariance to allow custom types

briancavalier:
It seems like this should be fine, since all the types line up visually.

```js
class NullSink {
  event (t: Time, x: any): void {}
  end (t: Time): void {}
  error (t: Time, e: Error): void {}
}

const nullSink: Sink<any> = new NullSink()
```

It isn't.  Flow considers object properties to be _invariant_ by default (See [Property Invariance here](https://flow.org/blog/2016/10/04/Property-Variance/). I'm still not sure exactly _why_ prototype properties violate the invariance rule, but they do:

```
src/split.js:16
 16: const nullSink: Sink<any> = new NullSink()
                                 ^^^^^^^^^^^^^^ NullSink. Covariant property `error` incompatible with invariant use in
 16: const nullSink: Sink<any> = new NullSink()
                     ^^^^^^^^^ object type
```

This PR adds explicit covariance for the properties of core types, to allow prototype properties satisfying core types, like the above, to type check.

screen shot 2017-07-11 at 7 19 43 am

### DIFF
--- a/packages/types/index.js.flow
+++ b/packages/types/index.js.flow
@@ -6,18 +6,18 @@
 export type Time = number
 
 export type Stream<A> = {
-  run: (Sink<A>, Scheduler) => Disposable
+  +run: (Sink<A>, Scheduler) => Disposable
 }
 
 export type Sink<A> = {
-  event: (Time, A) => void,
-  end: (Time) => void,
-  error: (Time, Error) => void,
+  +event: (Time, A) => void,
+  +end: (Time) => void,
+  +error: (Time, Error) => void,
 }
 
 // Interface of a resource that can be disposed
 export type Disposable = {
-  dispose: () => void
+  +dispose: () => void
 }
 
 // Delay time offset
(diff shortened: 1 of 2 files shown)


briancavalier:
I think there may be a better solution.  Instead of defining types with _properties_, it's possible to define them with _methods_.  For example, here is the Sink type as of `@most/types` 0.9.1, with properties, which fails for the case in the original PR description:

```js
export type Sink<A> = {
  event: (Time, A) => void,
  end: (Time) => void,
  error: (Time, Error) => void,
}
```

Here's the modified version, which, like the explicit covariant version, _works_ with the example in the PR description:

```js
export type Sink<A> = {
  event (Time, a: A): void,
  end (Time): void,
  error (Time, Error): void,
}
```

I think the latter matches most folks intuition, and would be much less of a WTF for someone who looks at the type defs.  I'll switch this PR to use that style.

packages/types/index.js.flow
```diff
@@ -6,18 +6,18 @@
...
 export type Stream<A> = {
-  run: (Sink<A>, Scheduler) => Disposable
+  run (Sink<A>, Scheduler): Disposable
 }
 
 export type Sink<A> = {
-  event: (Time, A) => void,
-  end: (Time) => void,
-  error: (Time, Error) => void,
+  event (Time, a: A): void,
```
briancavalier: Oops, a stray `a` parameter name!  Will fix, then merge.

PR #91, opened 2017-07-13
adds MulticastSource typings

TylorS:


packages/core/src/combinator/multicast.js
```diff
@@ -18,20 +18,20 @@ export class MulticastSource {
...
     if (n === 1) {
-      this._disposable = this.source.run(this, scheduler)
+      this.disposable = this.source.run(this, scheduler)
     }
     return disposeOnce(new MulticastDisposable(this, sink))
   }
 
-  _dispose () {
+  dispose () {
     const disposable = this._disposable
```
briancavalier: Here too 😄

packages/core/src/combinator/multicast.js
```diff
@@ -1,6 +1,6 @@
-import { append, remove, findIndex } from '@most/prelude'
+import { append, findIndex, remove } from '@most/prelude'
 import { disposeNone, disposeOnce } from '@most/disposable'
-import { tryEvent, tryEnd } from '../source/tryEvent'
+import { tryEnd, tryEvent } from '../source/tryEvent'
```
briancavalier: woot, import sorting!

TylorS: Oops, didn't even mean to do that, but I use `sort-imports` plugin in VSCode

briancavalier:
The failing tests are due to test fakes [here](https://github.com/mostjs/core/blob/master/packages/core/test/combinator/multicast-test.js#L161) and [here](https://github.com/mostjs/core/blob/master/packages/core/test/combinator/multicast-test.js#L183) still having `._dispose` rather than `.dispose`.  I wonder if it'd better to switch the tests to using real MulticastSources instead of maintaining fakes.  Thoughts?

packages/core/src/index.js.flow
```diff
@@ -178,6 +178,25 @@ declare export function throwError <E:Error> (e: Error): Stream<any>
 
 declare export function multicast <A> (s: Stream<A>): Stream<A>
 
+declare export class MulticastSource <A> {
+  source: Stream<A>;
+  disposable: Disposable;
+  sinks: Array<Sink<A>>;
```
briancavalier: Reorder properties to match TS defs.

TylorS:
Well we're already exporting `MulticastSource` so I think it'd be okay to switch the tests to use the real thing.

briancavalier:
> Well we're already exporting MulticastSource so I think it'd be okay to switch the tests to use the real thing.

Sounds good to me :+1:

PR #97, opened 2017-07-17
docs(api): zipArrayValues, withArrayValues, chain, and join

TylorS:


docs/api.rst
```diff
@@ -189,13 +189,70 @@ It allows you to maintain and update a "state" (aka feedback, aka seed for the n
...
+
+  withArrayValues :: [a] -> Stream b -> Stream a
+
+Creates a new stream by associating the time of event occurence from ``stream`` with the values of an array. The new stream 
+will end either when there are no more values in the array or the underlying stream has ended.::
+
+  array:                          [ 1, 2, 3 ]
+  stream:                         --x--x--x--x--x-->
+  withArrayValues(array, stream): --1--2--3|
+..
```
briancavalier: extra `..` here

docs/api.rst
```diff
@@ -189,13 +189,70 @@ It allows you to maintain and update a "state" (aka feedback, aka seed for the n
...
-chain :: (a -> Stream b) -> Stream a -> Stream b
+.. code-block:: haskell
+
+  zipArrayValues :: ((a, b) -> c) -> [a] -> Stream b -> Stream c
+
+Creates a new stream by applying a function ``((a, b) -> c)`` to the latest value of ``Stream b`` with an increasing index of ``[a]`` until ``Stream b`` ends or there are no more values in ``[a]``.::
+
+  stream:                             --10---10---10---10---10--->
+  array:                              [ 1, 2, 3 ]
+  zipArrayValues(add, array, stream): --11---12---13|
```
briancavalier: Love this example ❤️

docs/api.rst
```diff
@@ -189,13 +189,70 @@ It allows you to maintain and update a "state" (aka feedback, aka seed for the n
...
+.. code-block:: haskell
+
+  withArrayValues :: [a] -> Stream b -> Stream a
+
+Creates a new stream by associating the time of event occurence from ``stream`` with the values of an array. The new stream 
+will end either when there are no more values in the array or the underlying stream has ended.::
+
+  array:                          [ 1, 2, 3 ]
+  stream:                         --x--x--x--x--x-->
+  withArrayValues(array, stream): --1--2--3|
```
briancavalier: love this one too.

docs/api.rst
```diff
@@ -189,13 +189,70 @@ It allows you to maintain and update a "state" (aka feedback, aka seed for the n
...
-withArrayValues :: [a] -> Stream b -> Stream a
+zipArrayValues
+^^^^^^^^^^^^^^
 
-chain :: (a -> Stream b) -> Stream a -> Stream b
+.. code-block:: haskell
+
+  zipArrayValues :: ((a, b) -> c) -> [a] -> Stream b -> Stream c
+
+Creates a new stream by applying a function ``((a, b) -> c)`` to the latest value of ``Stream b`` with an increasing index of ``[a]`` until ``Stream b`` ends or there are no more values in ``[a]``.::
```
briancavalier: I've been trying to avoid starting every combinator description with "Create a new stream ..." (for sources, i.e. constructors, I think I ended up keeping it).  I feel like our theme of FP/immutability is strong enough that we don't need to say it.  Does that seem ok?

If so, how about something like:

"Apply a function to each event and the array value at the respective index."

Then maybe expand on the short description below the ascii diagram with the additional info about the end-of-stream.  I wonder if there is a way to say it somewhat declaratively, i.e. stating the proportion of the stream, rather than talking about "will end when" kind of things. (I know there are other places in the doc where I've written in the "ends when" style ... might try to do a pass and switch those if we feel the declarative style is the way we want to go)

Here's an attempt:

"The resulting stream will contain the same number of events as the input stream, or `array.length` events, whichever is less."

Thoughts?

docs/api.rst
```diff
@@ -189,13 +189,70 @@ It allows you to maintain and update a "state" (aka feedback, aka seed for the n
...
+
+withArrayValues
+^^^^^^^^^^^^^^^
+
+.. code-block:: haskell
+
+  withArrayValues :: [a] -> Stream b -> Stream a
+
+Creates a new stream by associating the time of event occurence from ``stream`` with the values of an array. The new stream 
+will end either when there are no more values in the array or the underlying stream has ended.::
```
briancavalier: Depending on where we end up with `zipArrayValues`, we can synchronize the style of `withArrayValues` to match (like you did here :+1:)

docs/api.rst
TylorS: I agree, I find the `creates a new stream ...` to be verbose as well. I was just taking lead from some of the other documentation. 

I also agree with your concerns about "ends when ..." too 👍

docs/api.rst
TylorS: I like the way it lines up

docs/api.rst
TylorS: Nice catch!

PR #101, opened 2017-07-25
docs(api): add documentation for tap and ap

TylorS:


docs/api.rst
```diff
@@ -171,20 +171,45 @@ Replace each event value with x.::
...
+
 ap
-^^
+^^^
 
 .. code-block:: haskell
 
   ap :: Stream (a -> b) -> Stream a -> Stream b
 
+Apply the latest function in a stream of functions to the latest value in stream.
```
briancavalier: The end of the sentence, "latest value in stream", seems like a copypasta artifact.  I wonder if it works to try to make the second half parallel to the first, like:

"Apply the latest function in a stream of functions to the latest value in a stream of values."

Does that read ok?

docs/api.rst
```diff
@@ -171,20 +171,45 @@ Replace each event value with x.::
...
+
+.. code-block:: javascript
+
+  streamOfFunctions:              --f---------g--------h------>
+  stream:                         -a-------b-------c-------d-->
+  ap(stream, streamOfFunctions.): --fa-----fb-gb---gc--hc--hd->
```
briancavalier: This one is really hard to draw an ascii diagram for.  Do you think the `fa`, `fb`, etc stuff will make sense to readers?  I know it's what's in the most.js docs, but should we try to come up with something better, or do you think it's ok?

docs/api.rst
Frikki: Does it matter if it is a "stream of values"? I mean, couldn’t it be a stream of anything? How about:

"Apply the latest emitted function in a stream of functions to the latest emitted value of another stream."

docs/api.rst
Frikki:
```
streamOfFunctions:              --f-----------g---------h--------->
stream:                         -a-------b---------c---------d---->
ap(stream, streamOfFunctions.): --f(a)---f(b)-g(b)-g(c)-h(c)-h(d)->
```

docs/api.rst
briancavalier: @frikki I like it :+1:

We've been trending away from the word emitted, but imho, your suggestion still sounds good after omitting it:

"Apply the latest function in a stream of functions to the latest value of another stream."

PR #103, opened 2017-07-27
docs(api): minor fixes

davidchase:
Just "style" changes/fixes and for `throwError` i switched the code block with the description to be consistent with the rest

I think `sample` may need some docs, i can add it to here or send in another PR


briancavalier:
> I think sample may need some docs, i can add it to here or send in another PR

That'd be awesome, thanks!  A separate PR sounds good.

PR #104, opened 2017-07-27
docs(api): add sample docs

davidchase:


docs/api.rst
```diff
@@ -406,7 +406,25 @@ Array form of :ref:`zip`.  Apply a function to corresponding events from all the
...
+  
+  s1:                          -1-----2-----3->
+  s2:                          -1---2---3---4->
+  sampler:                     -a-a-a-a-a-a-a->
+  sampler.sample(sum, s1, s2): -2-2-3-4-5-5-7->
+
+  s1:                          -1----2----3->
+  s2:                          -1-2-3-4-5-6->
+  sampler:                     -a--a--a--a-->
+  sampler.sample(sum, s1, s2): -2--3--6--7-->
```
briancavalier: These examples will need to be reworked, since `@most/core`'s `sample` takes only 2 streams: a sampler and a samplee.  We also need to convert from methods to functions.  So maybe like:

```
s1:                       -1--2--3--4--5->
sampler:                  -1-----2-----3->
sample(sum, sampler, s1): -2-----5-----8->

s1:                       -1-----2-----3->
sampler:                  -1--2--3--4--5->
sample(sum, sampler, s1): -2--3--5--6--8->

File: docs/api.rst [hunk 2]
```diff
@@ -406,7 +406,25 @@ Array form of :ref:`zip`.  Apply a function to corresponding events from all the
...
   zipArray(add3, [s1, s2, s3]): --3--6--9->
 
-sample :: ((a, b) -> c) -> Stream a -> Stream b -> Stream c
+sample
+^^^^^^
+.. code-block:: haskell 
+
+  sample :: ((a, b) -> c) -> Stream a -> Stream b -> Stream c
+
+Create a new stream by combining sampled values from many input streams.::
```
briancavalier: Hmmm, need some rewording here, since we only have a sampler and samplee, rather than "many input streams".  How about:

```
Sample one stream using another, and combine the values by applying a function.::
```

I don't love that wording, but maybe it'll inspire someone else to come up with something  better :)

docs/api.rst
davidchase: really bad copypasta :grimacing:

docs/api.rst
davidchase: 🤔 
```
Produce a behavior like stream by sampling another stream, combining the values by applying the given function
```

TylorS:
```
Applies a function `f`  with the latest values from 2 streams using one stream as the event's time occurrence.
```

Maybe something like?

Frikki: Sometimes we attempt to convey a lot of information in just one sentence, but perhaps splitting it up into a more natural language of more sentences will help.

briancavalier: @Frikki That's a great point.  I agree: trying to describe in 1 sentence is a good goal, but we shouldn't force it if 2-3 sentences makes for a better explanation.

On that note, how about something like:

> For each event in a _sampler_ stream, apply a function to combine it with the most recent event in another stream.  The resulting stream will contain the same number of events as the _sampler_ stream.

I also thought of this, which may be a more compact way to say the first sentence above:

> Combine each event in one stream with the most recent event in another.

Maybe we could also point out the difference between `sample` and `combine`.

Do either of those seem like a good direction?  Do they spark other ideas?

TylorS:
```
For each event in a sampler stream, apply a function to combine it with the most recent event in another stream. The resulting stream will contain the same number of events as the sampler stream.
```

👍 from me

Frikki: Same as @TylorS .

PR #116, opened 2017-08-28
chore(core): switch to prettier-standard

davidchase:
**WIP** 💥 
I think is a good place to have a discussion let me know what you guys think

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,4 +1,12 @@
 {
-  "parser": "babel-eslint",
-  "extends": "@most/eslint-config-most"
+  "parserOptions": {
+    "ecmaFeatures": {
+      "experimentalObjectRestSpread": false,
+      "jsx": false
+    }
+  },
+  "rules": {
+    "arrow-parens": [2, "as-needed"],
+    "complexity": [2, 3]
+  }
 }
--- a/packages/core/.eslintrc
+++ b/packages/core/.eslintrc
@@ -1,4 +1,12 @@
 {
-  "parser": "babel-eslint",
-  "extends": "@most/eslint-config-most"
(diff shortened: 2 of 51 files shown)


briancavalier:
Cool, thanks for getting this started!

From [the README](https://github.com/sheerun/prettier-standard#‿-prettier-standard---):

> While standard is a linter, prettier-standard is a formatter

The way I understand it is that prettier-standard is _only_ a formatter: it's prettier with a curated set of rules that match formatting expected by the standard linter.  So, I think we'd want to use both prettier-standard and standard (or eslint-prettier + estlint + eslint-config-standard?), since prettier-standard isn't a linter, and won't detect, for example, unused variables, and other mistakes.

Is that right?

davidchase:
Yes it seems you are correct it's not that obvious since to me if you are using prettier and standard I would expect both linting and formatting not only one that's a shame

briancavalier:
Yeah, I think the name "prettier-standard" fooled me, too.  I also thought it meant "prettier _and_ standard", but it seems to mean "prettier configured to use formatting rules that are compatible with standard".

It might not be bad to do something like `prettier-standard && standard`.  If they're both fast enough, that might be just fine.

The more manual approach of using prettier-eslint + eslint might also be a good way to go.

Frikki:
Separation of concerns, my dear friends.

davidchase:
I will try out a few things it would be nice to have a simple instal package that brings both without config hassles it shouldn't be this messy

TrySound:
Does original [prettier](https://github.com/prettier/prettier) with a pair of options look different?

davidchase:
closing in favor of #121

PR #117, opened 2017-08-29
feat(scheduler): Add currentTime free function for reading scheduler time

briancavalier:
This adds `currentTime :: Scheduler -> Time` as the function API to getting the current Time from a Scheduler.  It also deprecates `scheduler.now` in favor of `scheduler.currentTime`.  I'm not sure that's really necessary, but it could be confusing for the method to be named `now` and the function to be named `currentTime`.  I'm open to keeping the `scheduler.now` method if that seems better.

Of course, Timer and Clock also still have `now` methods, but it will be rare for anyone to interact with those, and when they do, they should really know what they're doing 😄


davidchase:
🐑 it real hard

PR #120, opened 2017-08-31
feat(core): Add runStream as function from of stream.run

briancavalier:
Add `runStream` as the function form of `stream.run`.

### Open questions

Argument ordering.  It's not clear what the "right" argument ordering is.  Since the subject of the function is a Stream, the obvious choice is to make it last.  After that, though, it's less clear.

Consider that we have two other, similar functions, and it might be nice to have some coherent criteria for ordering the arguments across `runStream` and these:

- [runEffects](https://github.com/mostjs/core/blob/master/packages/core/src/runEffects.js#L5) - `runEffects :: Stream -> Scheduler`, and
- [runWithLocalTime](https://github.com/mostjs/core/blob/master/packages/core/src/scheduler/runWithLocalTime.js#L11) - `runWithLocalTime :: Offset -> Stream a -> Sink a -> Scheduler`
    - I believe my thinking here was that it made sense for the last two args to mimic the argument ordering of `stream.run`, i.e. sink, scheduler.  Maybe it makes sense to move the Stream to the last arg to end up with `Offset -> Sink a -> Scheduler -> Stream a`, which is similar to `runStream` with the extra `Offset` arg prepended.

Thoughts and suggestions welcome!

### Todo

- [x] docs


davidchase:
I like stream last approach, in a similar to data last... given you can make a simple sink and have a defaultScheduler but not yet know what the stream is going to be entirely. If that makes any sense 😄

briancavalier:
Ok, I found a coherent argument ordering for runStream and runStreamWithLocalTime (newly renamed from runWithLocalTime) that I think makes sense.  I'm not really happy with the name runStreamWithLocalTime, though.  It'd be nice to find something shorter and at least as clear.

Ideas?

I'm inclined to keep runEffects' current "flipped" argument order (i.e. scheduler last, `Stream a -> Scheduler`).  I'm having trouble coming up with compelling reasons to justify that other than trying to provide some guidance about calling runEffects once and at the top level, instead of partially applying a Scheduler first and then scattering calls to runEffects around a codebase.  Obviously someone can just flip(runEffects), but at least they have to make that conscious decision.

davidchase:
why not just leave it `runWithLocalTime` and `run` or `runStream` ?

briancavalier:
My thinking is that the name `run` is a little less helpful than `runStream`, but maybe it's ok, since the types make it obvious.

The "WithLocalTime" suffix, while very explicit, seems a bit wordy.  I wonder if there's a way to say the same thing in less than 3 camel humps, like maybe "WithOrigin" or just "At": runWithOrigin, runAt, or runStreamWithOrigin, runStreamAt.  Or maybe "Relative": runRelative or runStreamRelative.

Do any of those seem appealing or give you other ideas?

I'm not totally against runWithLocalTime or runStreamWithLocalTime since they will tend to be used less often than runEffects, but if we _could_ find something more concise, that'd be great.

davidchase:
I like run because like you mention types and the fact this is a stream library but maybe that's me with that said I like runWithOrigin or runWithTime if you really want make it obvious .. though for some reason it almost makes sense to make it runWithEffects ?
So something like
run
runWithEffects
runWithOrigin or runWithTime

Thoughts ?

packages/core/src/runStream.js
```diff
@@ -1,4 +1,24 @@
...
 
+// Run a Stream, sending all its events to the
+// provided Sink.
 export const runStream = curry3((sink, scheduler, stream) =>
     stream.run(sink, scheduler))
+
+// Run a stream with its own localized clock
+// This transforms time from the provided scheduler's clock to a stream-local
+// clock (which starts at 0), and then *back* to the scheduler's clock before
+// propagating events to sink.  IOW, stream.run will see local times, and sink
```
Frikki: "IOW" is _slang_ for in other words. Google sees it as Isle of Wright. Wikipedia lists:

- Isle of Wight (frequently as IoW)
- The German abbreviation for the Leibniz Institute for Baltic Sea Research
- ISO 639-3 code of the Iowa-Oto language

https://en.wikipedia.org/wiki/IOW

I suggest that you only use official acronyms, so I think you should change it to "In other words, ..."

docs/api.rst
```diff
@@ -187,7 +187,29 @@ runEffects
...
+.. _runStreamWithLocalTime:
+
+runStreamWithLocalTime
+``````````````````````
+
+.. code-block:: haskell
+
+  runStreamWithLocalTime :: Sink a -> Scheduler -> Time -> Stream a -> void
+
+Run a :ref:`Stream`, sending all events to the provided :ref:`Sink`.  The Stream will have localized :ref:`Time` values, whose origin (i.e. time 0) is at the specified Time on the provided :ref:`Scheduler`.
```
Frikki: Comma after "i.e.". So:

... whose origin (i.e., time 0)

The same goes for e.g.

Note that the comma after the abbreviations is American style. The British style omits the comma.

briancavalier:
> though for some reason it almost makes sense to make it runWithEffects

I'm not sure I understand.  We started with `runEffects` as the primary application-level API for activating a stream graph, and this PR adds two functions, currently named `runStream` and `runStreamWithLocalTime`, primarily to be used when authoring custom stream types. Are you suggesting we rename that to `runWithEffects`, or that we rename the `runStream` in this PR to `runWithEffects`?

One tricky thing now is that all 3 of these functions, `runEffects`, `runStream`, and `runStreamWithLocalTime`, run effects.  That is, they all activate a stream and begin consuming all its events and performing any side-effects.  They differ in their parameters and return values.

I can even imagine other functions for activating a stream:

```hs
-- Provide a handler for error, and a handler for end
cleverName :: (Error -> void) -> (() -> void) -> Stream a -> Scheduler -> void

-- Provide a handler for error only
cleverName2 :: (Error -> void) -> Stream a -> Scheduler -> void
```

Which maybe opens the can of worms about whether returning a promise is even worthwhile.  Given the first variant above, someone could easily do this to get themselves a promise:

```js
new Promise((resolve, reject) => cleverName(resolve, reject, stream, scheduler))
```

briancavalier:
A few thoughts on short names (even though I don't think short is a necessity here since these functions won't be called all that often):

run
runAt

runStream
runStreamAt

run
runWithTime
(or runWithOrigin)

packages/core/src/runStream.js
briancavalier: Thanks! Fixed in 034e260

docs/api.rst
briancavalier: Fixed in 034e260

Frikki:
`runLocalized`

briancavalier:
See 46a6511, which splits stream clock localization into its own withLocalTime combinator.  Instead of needing 3 run-like functions, it means we only need 2: the existing runEffects(), and run(), where run is just the trivial function version of stream.run: `run :: Sink a -> Scheduler -> Stream a -> Disposable`.

briancavalier:
If we like this direction, then there are a couple remaining todos:

- [ ] Add unit tests for withLocalTime
- [ ] Add docs for run and withLocalTime

davidchase:
Love it 😍

packages/core/src/index.js.flow
```diff
@@ -6,7 +6,22 @@ export type SeedValue <S, A> = {
...
+
+declare export function runStreamWithLocalTime <A> (sink: Sink<A>, scheduler: Scheduler, origin: Time, s: Stream<A>): Disposable
+declare export function runStreamWithLocalTime <A> (sink: Sink<A>): (scheduler: Scheduler, origin: Time, s: Stream<A>) => Disposable
+declare export function runStreamWithLocalTime <A> (sink: Sink<A>): (scheduler: Scheduler) => (origin: Time, s: Stream<A>) => Disposable
+declare export function runStreamWithLocalTime <A> (sink: Sink<A>): (scheduler: Scheduler, origin: Time) => (s: Stream<A>) => Disposable
+declare export function runStreamWithLocalTime <A> (sink: Sink<A>): (scheduler: Scheduler) => (origin: Time) => (s: Stream<A>) => Disposable
+declare export function runStreamWithLocalTime <A> (sink: Sink<A>, scheduler: Scheduler): (origin: Time, s: Stream<A>) => Disposable
+declare export function runStreamWithLocalTime <A> (sink: Sink<A>, scheduler: Scheduler): (origin: Time) => (s: Stream<A>) => Disposable
+declare export function runStreamWithLocalTime <A> (sink: Sink<A>, scheduler: Scheduler, origin: Time): (s: Stream<A>) => Disposable
```
briancavalier: Whoops!  I need to update the TS and Flow type definitions

briancavalier:
- [ ] Update TS and Flow type defs

docs/api.rst
```diff
@@ -942,6 +956,22 @@ Timeshift a :ref:`Stream` by *n* milliseconds. ::
...
+.. _withLocalTime:
+
+withLocalTime
+`````````````
+
+.. code-block:: haskell
+
+  withLocalTime :: Offset -> Stream a -> Stream a
+
+Create a Stream with localized :ref:`Time` values, whose origin (i.e., time 0) is at the specified Time on the :ref:`Scheduler` used to the Stream is observed with :ref:`runEffects` or :ref:`run`.
```
davidchase: > Scheduler used to the Stream

something maybe missing here ?

Frikki: Yes. That is not a complete sentence.

docs/api.rst
```diff
@@ -942,6 +956,22 @@ Timeshift a :ref:`Stream` by *n* milliseconds. ::
...
+
+When implementing custom higher-order :ref:`Stream` combinators, such as :ref:`chain`, you should use ``withLocalTime`` to localize "inner" Streams before running them.
```
Frikki: How is `chain` a custom combinator that the user implements?

docs/api.rst
Frikki: In the code comment for ` withLocalTime`, the following text is found:

> Create a stream with its own local clock
> This transforms time from the provided scheduler's clock to a stream-local clock (which starts at 0), and then *back* to the scheduler's clock before propagating events to sink.  In other words, upstream sources will see local times, and downstream sinks will see non-local (original) times.

packages/core/src/combinator/withLocalTime.js
```diff
@@ -1,15 +1,24 @@
-/** @license MIT License (c) copyright 2010-2017 original author or authors */
-
 import RelativeSink from '../sink/RelativeSink'
 import { schedulerRelativeTo } from '@most/scheduler'
 
-// Run a stream with its own localized clock
+// Create a stream with its own local clock
```
Frikki: > ... with its own local clock

Do we also have a clock type? I know there is  `Time`.

davidchase: dah we do http://mostcore.readthedocs.io/en/latest/api.html#clock

docs/api.rst
briancavalier: "such as" is referring to chain as a higher-order combinator.  Would you suggest removing the word "custom"?

docs/api.rst
briancavalier: I think in this case, the code comment provides some extra detail on how withLocalTime works, which may not be appropriate for user api docs.  On the other hand, the last sentence in the comment may be useful in the docs.

From your comment, I'm not sure what you're suggesting.  Could you clarify?  Thanks!

packages/core/src/combinator/withLocalTime.js
Frikki: Mmmokay

briancavalier:
🎉

PR #122, opened 2017-09-01
chore(linting): only add standard for now

davidchase:
third time the charm 😄


packages/core/test/combinator/multicast-test.js
```diff
@@ -51,7 +51,7 @@ describe('multicast', () => {
   })
 })
 
-const sentinel = { value: sentinel }
+const sentinel = { value: 'sentinel' }
```
briancavalier: 😵

davidchase: its ok have a 🥃

briancavalier:
This is awesome, @davidchase. Thanks for trying out several alternatives!

PR #123, opened 2017-09-02
fix(types): propagateTask; decouple value type from sink type

axefrog:
Without this change, propagateTask can't be used to pass a wrapper object as the value, with the internal values then unwrapped by the task runner for delivery to the sink.

### DIFF
--- a/packages/core/type-definitions/PropagateTask.d.ts
+++ b/packages/core/type-definitions/PropagateTask.d.ts
@@ -1,9 +1,9 @@
 import { Sink, Task } from '@most/types';
 
-export function propagateTask<T>(run: PropagateTaskRun<T>, value: T, sink: Sink<T>): PropagateTask<T>;
-export function propagateTask<T>(run: PropagateTaskRun<T>, value: T): (sink: Sink<T>) => PropagateTask<T>;
-export function propagateTask<T>(run: PropagateTaskRun<T>): (value: T, sink: Sink<T>) => PropagateTask<T>;
-export function propagateTask<T>(run: PropagateTaskRun<T>): (value: T) => (sink: Sink<T>) => PropagateTask<T>;
+export function propagateTask<A, B = A>(run: PropagateTaskRun<A, B>, value: A, sink: Sink<B>): PropagateTask<A, B>;
+export function propagateTask<A, B = A>(run: PropagateTaskRun<A, B>, value: A): (sink: Sink<B>) => PropagateTask<A, B>;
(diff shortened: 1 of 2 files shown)


TylorS:
This looks good to me 👍

packages/core/type-definitions/PropagateTask.d.ts
```diff
@@ -1,9 +1,9 @@
 import { Sink, Task } from '@most/types';
 
-export function propagateTask<T>(run: PropagateTaskRun<T>, value: T, sink: Sink<T>): PropagateTask<T>;
-export function propagateTask<T>(run: PropagateTaskRun<T>, value: T): (sink: Sink<T>) => PropagateTask<T>;
-export function propagateTask<T>(run: PropagateTaskRun<T>): (value: T, sink: Sink<T>) => PropagateTask<T>;
-export function propagateTask<T>(run: PropagateTaskRun<T>): (value: T) => (sink: Sink<T>) => PropagateTask<T>;
+export function propagateTask<A, B = A>(run: PropagateTaskRun<A, B>, value: A, sink: Sink<B>): PropagateTask<A, B>;
```
briancavalier: Can you help me understand this `B = A` constraint?  If A and B are always the same type, then I'm not seeing how this is different from a single type parameter

TylorS: It's a type default. So if you only supply one type `propagateTask<number>(...)` it will assume `B` is of type `number`. Otherwise `propagateTask<number, string>(..)` will set A to number and B to string. It actually makes this a non-breaking change afaik

axefrog: @TylorS is correct; it preserves the default expectation while allowing for variance when needed.

briancavalier: Ohhhhh, I thought it was a constraint! Thanks! I don't mind a breaking change, but it's pretty cool that this makes it backward compatible.

I don't think flow has a notion of a type default, but we should update the flow refs to match without a default. Could you include that in this PR as well, @axefrog?

axefrog: I could... I'd need to disclaim that I have no familiarity with flow, so you'd need to double check that I've done it correctly.

briancavalier: No worries.  We're happy to help/review/etc.  The change should be [basically equivalent here](https://github.com/mostjs/core/blob/master/packages/core/src/index.js.flow#L200-L211) except that flow doesn't support default types (unless I've missed it, which is also possible!).

axefrog: Sorry for the delay. PR updated.

PR #128, opened 2017-09-07
Correct grammar, typos, spelling, and formatting

Frikki:
Unify style and language while correcting various grammar, typos, and spelling.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -224,7 +224,7 @@ empty
 
   empty :: () -> Stream *
 
-Create a stream containing no events, which ends immediately.
+Create a stream containing no events, which ends immediately. ::
 
   empty(): |
 
@@ -237,7 +237,7 @@ never
 
   never :: () -> Stream *
 
-Create a stream containing no events, which never ends.
+Create a stream containing no events, which never ends. ::
 
   never(): ---->
 
@@ -250,7 +250,7 @@ now
 
   now :: a -> Stream a
 
-Create a stream containing a single event at time 0.
+Create a stream containing a single event at time 0. ::
 
   now(x): x|
 
@@ -263,7 +263,7 @@ at
 
   at :: Time -> a -> Stream a
 
-Create a stream containing a single event at a specific time.
+Create a stream containing a single event at a specific time. ::
 
   at(3, x): --x|
 
@@ -279,7 +279,7 @@ throwError
 Create a stream that fails at time 0 with the provided Error.
 
 This can be useful for functions that need to return a stream and also need 
-to propagate an error.
+to propagate an error. ::
 
   throwError(X): X
 
@@ -302,7 +302,7 @@ Prepend an event at time 0.
 
 Note that ``startWith`` *does not* delay other events.  If ``stream`` already 
 contains an event at time 0, then ``startWith`` simply adds another event at 
-time 0—the two will be simultaneous, but ordered.  For example::
+time 0—the two will be simultaneous, but ordered. For example::
 
   stream:                a-b-c-d->
   startWith(x, stream): xa-b-c-d->
@@ -318,7 +318,7 @@ continueWith
 
   continueWith :: (() -> Stream a) -> Stream a -> Stream a
 
-Replace the end of a stream with another stream.
+Replace the end of a stream with another stream. ::
 
   s:                  -a-b-c-d|
   f(): 		                    -1-2-3-4-5->
@@ -335,7 +335,7 @@ map
 
   map :: (a -> b) -> Stream a -> Stream b
 
-Apply a function to each event value.
+Apply a function to each event value. ::
 
   stream:        -a-b-c-d->
   stream.map(f): -f(a)-f(b)-f(c)-f(d)->
@@ -353,7 +353,7 @@ constant
 
   constant :: a -> Stream * -> Stream a
 
-Replace each event value with x.
+Replace each event value with x. ::
 
   stream:              -a-b-c-d->
   constant(x, stream): -x-x-x-x->
@@ -412,7 +412,7 @@ scan
 
   scan :: (b -> a -> b) -> b -> Stream a -> Stream b
 
-Incrementally accumulate results, starting with the provided initial value.
+Incrementally accumulate results, starting with the provided initial value. ::
 
   stream:                           -1-2-3->
   scan((x, y) => x + y, 0, stream): 01-3-6->
@@ -461,7 +461,7 @@ zipArrayValues
 
   zipArrayValues :: ((a, b) -> c) -> [a] -> Stream b -> Stream c
 
-Apply a function to the latest event and the array value at the respective index.
+Apply a function to the latest event and the array value at the respective index. ::
 
   stream:                             --10---10---10---10---10--->
   array:                              [ 1, 2, 3 ]
@@ -479,7 +479,7 @@ withArrayValues
 
   withArrayValues :: [a] -> Stream b -> Stream a
 
-Replace each event value with the array value at the respective index.
+Replace each event value with the array value at the respective index. ::
 
   array:                          [ 1, 2, 3 ]
   stream:                         --x--x--x--x--x-->
@@ -501,7 +501,7 @@ switchLatest
   switchLatest :: Stream (Stream a) -> Stream a
 
 Given a higher-order stream, return a new stream that adopts the behavior of
-(i.e., emits the events of) the most recent inner stream.
+(i.e., emits the events of) the most recent inner stream. ::
 
   s:                    -a-b-c-d-e-f->
   t:                    -1-2-3-4-5-6->
@@ -518,7 +518,7 @@ join
   join :: Stream (Stream a) -> Stream a
 
 Given a higher-order stream, return a new stream that merges all the inner 
-streams as they arrive.
+streams as they arrive. ::
 
   s:             ---a---b---c---d-->
   t:             -1--2--3--4--5--6->
@@ -535,7 +535,7 @@ chain
   chain :: (a -> Stream b) -> Stream a -> Stream b
 
 Transform each event in ``stream`` into a new stream, and then merge each into 
-the resulting stream. Note that ``f`` must return a stream.
+the resulting stream. Note that ``f`` must return a stream. ::
 
   stream:            -a----b----c|
   f(a):               1--2--3|
@@ -556,7 +556,7 @@ Transform each event in ``stream`` into a new stream, and then concatenate each 
 end of the resulting stream. Note that ``f`` must return a stream.
 
 The mapping function ``f`` is applied lazily. That is, ``f`` is called only once 
-it is time to concatenate a new stream.
+it is time to concatenate a new stream. ::
 
   stream:                -a----b----c|
   f(a):                   1--2--3|
@@ -580,7 +580,7 @@ mergeConcurrently
 Given a higher-order stream, return a new stream that merges inner streams as
 they arrive up to the specified concurrency. Once concurrency number of streams
 are being merged, newly arriving streams will be merged after an existing one
-ends.
+ends. ::
 
   s:                            --a--b--c--d--e-->
   t:                            --x------y|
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -201,7 +224,7 @@ empty
 
   empty :: () -> Stream *
 
-Create a stream containing no events, which ends immediately.::
```
davidchase: I think removing the `::` affects the way the following lines are formatted, no?

Frikki: You’re correct, @davidchase. Two colons indicate that the following indented or quoted text is a literal block. I will add them back.

docs/api.rst
```diff
@@ -271,12 +295,14 @@ startWith
 
   startWith :: a -> Stream a -> Stream a
 
-Prepend an event at time 0.::
+Prepend an event at time 0.
```
davidchase: I think this one is also missing `::`

Frikki: Good catch! Thanks.

docs/api.rst
```diff
@@ -153,7 +174,7 @@ Task
     error:: Time -> Error -> void
   }
 
-A Task is any unit of work that can be scheduled for execution on a Scheduler.
+A Task is any unit of work that can be scheduled for execution in a Scheduler.
```
briancavalier: "in" still sounds odd to me.  What about "with" or "using"?  I guess there are at least a couple ways to look at a Scheduler: 1) as a container of Tasks, 2) as an actor whose job it is to schedule tasks (and scheduled tasks "reside" in some other container).

2 may be more technically correct, since Timeline is the ultimate container.  If that's the case, then "with" or "using" may be more appropriate.

Thoughts?

docs/api.rst
```diff
@@ -166,7 +187,9 @@ ScheduledTask
     error :: Error -> void
   }
 
-A Scheduled Task represents a :ref:`Task` which has been scheduled on a particular :ref:`Scheduler`.  A ``ScheduledTask``'s ``dispose`` method will cancel the Task on the Scheduler on which it was scheduled.
+A Scheduled Task represents a :ref:`Task` which has been scheduled in a particular
```
briancavalier: Similarly here, re: "in".  Whatever we decide above, we can apply here.

docs/api.rst
briancavalier: One thing that's weird about the double-colon is that it ends up being rendered as a single colon.  So, we end up with perdiod-colon (like: "... which ends immediately.:"), which looks weird.  Not sure what to do about that ... remove the period in those cases?

docs/api.rst
```diff
@@ -345,16 +371,17 @@ tap
...
 
   stream:         -a-b-c-d->
   tap(f, stream): -a-b-c-d->
 
-For each event in stream, f is called, but the value of its result is ignored.
-If f fails (ie throws), then the returned stream will also fail. The stream
-returned by tap will contain the same events as the original stream.
+For each event in ``stream``, ``f`` is called, but the value of its result 
+is ignored. If ``f`` fails (i.e., throws an error), then the returned stream 
+will also fail. The stream returned by `tap` will contain the same events
```
briancavalier: `tap` needs double backticks, since this is RST. My markdown muscle memory still types single backticks all the time :/

docs/api.rst
```diff
@@ -430,14 +461,14 @@ zipArrayValues
...
-Apply a function to the latest event and the array value at the respective index.::
+Apply a function to the latest event and the array value at the respective index. ::
 
   stream:                             --10---10---10---10---10--->
   array:                              [ 1, 2, 3 ]
   zipArrayValues(add, array, stream): --11---12---13|
 
 The resulting stream will contain the same number of events as the input stream,
-or array.length events, whichever is less.
+or `array.length` events, whichever is less.
```
briancavalier: double backticks

docs/api.rst
```diff
@@ -448,14 +479,14 @@ withArrayValues
...
-Replace each event value with the array value at the respective index.::
+Replace each event value with the array value at the respective index. ::
 
   array:                          [ 1, 2, 3 ]
   stream:                         --x--x--x--x--x-->
   withArrayValues(array, stream): --1--2--3|
 
 The resulting stream will contain the same number of events as the input stream,
-or array.length events, whichever is less.
+or `array.length` events, whichever is less.
```
briancavalier: double backticks

docs/api.rst
```diff
@@ -519,11 +552,11 @@ concatMap
 
   concatMap :: (a -> Stream b) -> Stream a -> Stream b
 
-Transform each event in stream into a stream, and then concatenate it onto the
-end of the resulting stream. Note that f must return a stream.
+Transform each event in ``stream`` into a new stream, and then concatenate each onto the
```
briancavalier: To me, "new" implies that `f` must construct a stream to return, but any stream will do.  For example, it's valid to return the same `Stream b` for every `a`.  Granted, that's probably fairly rare, but it's allowed.

So, I think I prefer without "new".

Also, there's an opportunity to link to Stream:

```
Transform each event in ``stream`` into a :ref:`Stream`, and then ...
```

Overall, the docs aren't consistent in saying "stream" vs. "Stream" with link 😢 Laziness on my part, sorry!

docs/api.rst
```diff
@@ -547,19 +580,19 @@ mergeConcurrently
...
+ends. ::
 
   s:                            --a--b--c--d--e-->
   t:                            --x------y|
   u:                            -1--2--3--4--5--6>
   stream:                       -s--t--u--------->
   mergeConcurrently(2, stream): --a--b--cy4d-5e-6>
 
-Note that u is only merged after t ends, due to the concurrency level of 2.
+Note that ``u`` is only merged after ``t`` ends because of the concurrency level of 2.
```
briancavalier: TIL ["due to" vs "because of"](http://linguistech.ca/Tips+and+Tricks+-+Because+of+vs.+Due+to) :+1:

docs/api.rst
```diff
@@ -572,10 +605,10 @@ mergeMapConcurently
 
   mergeMapConcurently :: (a -> Stream b) -> int -> Stream a -> Stream b
 
-Lazily applies a function ``f`` to each event on a stream, merging them into the
+Lazily apply a function ``f`` to each event in a stream, merging them into the
 resulting stream at the specified concurrency. Once concurrency number of streams
 are being merged, newly arriving streams will be merged after an existing one
-ends.::
+ends. ::
```
briancavalier: Hmmm, this paragraph now mixes soft and hard returns.  I think we just need to pick one and go with it.  As mentioned above, I prefer relying on soft wrap, but won't die on a hill for it if there's a compelling reason to use hard returns everywhere.

docs/api.rst
```diff
@@ -589,8 +622,8 @@ Note that ``f(c)`` is only merged after ``f(a)`` ends.
 Also note that ``f`` will not get called with ``d`` until either ``f(b)`` or
 ``f(c)`` ends.
 
-To control concurrency, mergeMapConcurrently must maintain an internal queue of
-newly arrived streams. If new streams arrive faster than the concurrency level
+To control concurrency, ``mergeMapConcurrently`` must maintain an internal queue 
+of newly arrived streams. If new streams arrive faster than the concurrency level
```
briancavalier: This paragraph also mixes soft and hard returns now.

docs/api.rst
```diff
@@ -837,7 +888,7 @@ takeWhile
 
   takeWhile :: (a -> bool) -> Stream a -> Stream a
 
-Keep all events until predicate returns false, and discard the rest.::
+Keep all events until predicate returns ``false``, and discard the rest. ::
```
briancavalier: :+1: I like that you've made this explicitly boolean `false` (rather than "falsy"), to reinforce the strong typing mind-set.  How about we do the same above with `skipRepeatsWith`?

docs/api.rst
```diff
@@ -939,9 +992,9 @@ This is similar to :ref:`slice`, but uses time rather than indices to "slice" th
   // 2. ends at time = 6 seconds (1 second + 5 seconds)
   const timeWindow = at(1000, at(5000, null))
 
-  // 1. discard events for 1 second, then
+  // 1. Discard events for 1 second, then
   // 2. keep events for 5 more seconds, then
-  // 3. discard all subsequent events
+  // 3. discard all subsequent events.
```
briancavalier: :heart:

docs/api.rst
```diff
@@ -1313,7 +1396,7 @@ newPerformanceClock
 
   newPerformanceClock :: () -> Clock
 
-Create a new :ref:`Clock` using`performance.now`.
+Create a new :ref:`Clock` using `performance.now`.
```
briancavalier: Oops, sorry!  double backticks for `performance.now`

docs/api.rst
```diff
@@ -1324,7 +1407,7 @@ newHRTimeClock
 
   newHRTimeClock :: () -> Clock
 
-Create a new :ref:`Clock` using`process.hrtime`.
+Create a new :ref:`Clock` using `process.hrtime`.
```
briancavalier: Oops again.  Needs double backticks.

docs/api.rst
```diff
@@ -1335,7 +1418,9 @@ newDateClock
 
   newDateClock :: () -> Clock
 
-Create a new :ref:`Clock` using`Date.now`. Note that a Clock using `Date.now` is not guaranteed to be monotonic, and is subject to system clock changes, e.g. NTP can change your system clock!
+Create a new :ref:`Clock` using `Date.now`. Note that a Clock using `Date.now`
```
briancavalier: Oh man, my markdown muscle memory was in full swing: double backticks for these two `Date.now`s.

docs/api.rst
Frikki: @briancavalier, they will _only_ be rendered as a colon when they’re not preceded by a space.

docs/api.rst
```diff
@@ -15,15 +15,19 @@ Time
 
   type Time = number
 
-Time is a monotonic number. It represents the current time according to a Scheduler.  The default Scheduler uses ``performance.now`` in browsers, and ``process.hrtime`` (transformed to a `number`) in Node.
+Time is a monotonic number. It represents the current time according to 
+a Scheduler.  The default Scheduler uses ``performance.now`` in browsers and 
+``process.hrtime`` (transformed to a `number`) in Node.
```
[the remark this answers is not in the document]
Frikki: Yes, the lines run out of the screen otherwise. Assuming that soft-wrapping is available is like assuming a certain editor.

docs/api.rst
```diff
@@ -36,11 +40,20 @@ Stream
     run :: Sink a -> Scheduler -> Disposable
   }
 
-A Stream represents a view of events over time. It's ``run`` method arranges to propagate events to the provided Sink. Each stream has a local clock, defined by the provided :ref:`Scheduler`, which has methods for knowing the current time, and scheduling future tasks.
+A Stream represents a view of events over time. Its ``run`` method arranges to
```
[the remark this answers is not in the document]
Frikki: :+1:

docs/api.rst
```diff
@@ -55,9 +68,12 @@ Sink
     end :: Time -> void
   }
 
-A sink receives events, typically does something with them, such as transforming or filtering them, and then propagates them to another sink.
+A Sink receives events—typically it does something with them, such as
```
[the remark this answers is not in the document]
Frikki: They are em-dashes.

docs/api.rst
Frikki: I agree with 2. I used _in_ because I looked at a Scheduler as a mechanism.

docs/api.rst
Frikki: @briancavalier, I noticed that the lowercase stream is used when it refers to the literal block that follows.

docs/api.rst
Frikki: Done!

davidchase:
🐑  it

PR #129, opened 2017-09-08
chore(core): update some files to es6 classes

davidchase:


packages/core/src/Queue.js
```diff
@@ -4,58 +4,60 @@
 
 // Based on https://github.com/petkaantonov/deque
 
-export default function Queue (capPow2) {
-  this._capacity = capPow2 || 32
-  this._length = 0
-  this._head = 0
-}
+export default class Queue {
+  constructor(capPow2) {
```
Frikki: I think it’s cleaner and more visible to the client to have the default value set with the parameter:

```
constructor(capPow2 = 32)
```

briancavalier: Agreed.  @davidchase, could you change it in this PR?

packages/core/src/Queue.js
```diff
@@ -4,58 +4,60 @@
...
+    this._capacity = capPow2 || 32
+    this._length = 0
+    this._head = 0
+  }
 
-Queue.prototype.push = function (x) {
-  var len = this._length
-  this._checkCapacity(len + 1)
+  push(x) {
+    const len = this._length;
```
briancavalier: We've been infiltrated by semicolons 😱 (a few below, too)

davidchase: semi-🚚 FTW 😆

davidchase:
🍻

PR #133, opened 2017-09-16
feat(pipe): add pipe

nissoh:
commit includes pipe

pointless(pointfree) style example
````typescript
import { map, now } from '@most/core'
import { pipe } from '@most/prelude'

const hello =          x => x + 'hello'
const world =         x => x + 'world'
const addSpace =  x => x + ' '

const helloWorld = pipe(map(hello), pipe(map(addSpace), map(world)))

helloWorld(now('')) // 'hello world'|

````

### DIFF
--- a/packages/prelude/src/function.js
+++ b/packages/prelude/src/function.js
@@ -3,6 +3,9 @@
 // id :: a -> a
 export const id = x => x
 
+// pipe :: (a -> b) -> (b -> c) -> (a -> c)
+export const pipe = (f, g) => x => g(f(x))
+
 // compose :: (b -> c) -> (a -> b) -> (a -> c)
 export const compose = (f, g) => x => f(g(x))
 
--- a/packages/prelude/src/index.js.flow
+++ b/packages/prelude/src/index.js.flow
@@ -30,6 +30,8 @@ declare export function isArrayLike(x: any): boolean
 
 declare export function id<A>(A): A
 
+declare export function pipe <A, B, C>(f: (a: A) => B, g: (b: B) => C): (x: A) => C
+
 declare export function compose<A, B, C>(f: (B) => C, g: (A) => B): (A) => C
 
 declare export function apply<A, B>(f: (A) => B, A): B
--- a/packages/prelude/src/index.js.flow
+++ b/packages/prelude/src/index.js.flow
@@ -30,7 +30,7 @@ declare export function isArrayLike(x: any): boolean
 
 declare export function id<A>(A): A
 
-declare export function pipe <A, B, C>(f: (a: A) => B, g: (b: B) => C): (x: A) => C
+declare export function pipe <A, B, C>(f: (A) => B, g: (B) => C): (A) => C
 
 declare export function compose<A, B, C>(f: (B) => C, g: (A) => B): (A) => C
 
--- a/packages/prelude/test/function-test.js
+++ b/packages/prelude/test/function-test.js
@@ -13,7 +13,7 @@ describe('id', () => {
 })
 
 describe('pipe', () => {
-  it('pipe(f, g)(x) === f(g(x))', () => {
+  it('pipe(f, g)(x) === g(f(x))', () => {
     const fx = 'fx'
     const gx = 'gx'
     const x = 'x'
@@ -28,9 +28,9 @@ describe('pipe', () => {
 
 describe('compose', () => {
   it('compose(f, g)(x) === f(g(x))', () => {
-    const fx = '' + Math.random()
-    const gx = '' + Math.random()
-    const x = '' + Math.random()
+    const fx = 'fx'
+    const gx = 'gx'
+    const x = 'x'
 
     const f = x => x + fx
     const g = x => x + gx
--- a/packages/prelude/test/function-test.js
+++ b/packages/prelude/test/function-test.js
@@ -3,12 +3,26 @@
 import {describe, it} from 'mocha'
 import assert from 'assert'
 
-import { id, compose, apply, curry2, curry3, curry4 } from '../src/function'
+import { id, compose, pipe, apply, curry2, curry3, curry4 } from '../src/function'
 
 describe('id', () => {
   it('id(x) === x', () => {
     const x = {}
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by TrySound** [comment]
@briancavalier What about dynamic arguments? Flow will add `$Compose` and `$ComposeReverse` type utils in the next (or one of the next) versions. For now it could be typed with a fixed reloaded arguments.

**Comment by briancavalier** [comment]
@TrySound Do you have links to information about `$Compose` and `$ComposeReverse`, and when they will land?  I'd need to learn what they are before I can say anything about them.

Until there's a good solution for the types, I'd rather stick with 1-arg functions.

File: packages/prelude/src/index.js.flow [hunk 1]
```diff
@@ -30,6 +30,8 @@ declare export function isArrayLike(x: any): boolean
 
 declare export function id<A>(A): A
 
+declare export function pipe <A, B, C>(f: (a: A) => B, g: (b: B) => C): (x: A) => C
```
briancavalier: You can omit the parameter names inside the function parameters.  See `compose` below.

packages/prelude/type-definitions/index.d.ts
```diff
@@ -27,6 +27,8 @@ export function isArrayLike(x: any): boolean;
 
 export function id<A>(x: A): A;
 
+export function pipe <A, B, C>(f: (a: A) => B, g: (b: B) => C): (x: A) => C
```
briancavalier: Please add a semicolon to match the style in the rest of the file.  Thanks!

packages/prelude/test/function-test.js
```diff
@@ -12,6 +12,20 @@ describe('id', () => {
   })
 })
 
+describe('pipe', () => {
+  it('pipe(f, g)(x) === f(g(x))', () => {
+    const fx = 'fx'
+    const gx = 'gx'
+    const x = 'x'
```
briancavalier: Please match the style of the tests below by using `Math.random`, or update the tests below to match the style of this one.  Either is ok with me.

TrySound:
@briancavalier In the next release. Commit is merged.
https://github.com/facebook/flow/commit/882e5be2b4692dae6301e0d6a9f08c8d980a7fe2#diff-9b8fab691c00b9e5380b19ce882f3271R11

briancavalier:
Thanks for the link, @TrySound.  Let's keep an eye on them, and evaluate ways we might use them once they appear in the docs.

packages/prelude/src/function.js
```diff
@@ -3,6 +3,9 @@
 // id :: a -> a
 export const id = x => x
 
+// pipe :: (a -> b) -> (b -> c) -> (a -> c)
+export const pipe = (f, g) => x => g(f(x))
+
 // compose :: (b -> c) -> (a -> b) -> (a -> c)
 export const compose = (f, g) => x => f(g(x))
```
Frikki: Here’s another variant of `compose`, which handles more than just two functions:
```javascript
const compose = (...fns) => fns.reduce((f, g) => (...args) => f(g(...args)))

// Usage : compose functions right to left
// compose(minus8, add10, multiply10)(4) === 42
//
// The resulting function can accept as many arguments as the first function does
// compose(add2, multiply)(4, 10) === 42
```

What do you think, @briancavalier?

nissoh: @Frikki, that can work too, there may be a few caveats that comes to mind

- i think you will lose performance because of v8's deterministic inlining of functions
- current versions of typescript & flow(i assume) are bad at inferring variadic functions(`typescript`, possibly `flow`)

If you don't mind losing performance and/or type validation at runtime(flow), IDE(typescript vscode) and transpiration variadic approach is better then current compose/pipe


btw, variadic `compose` and `pipe` can be reused using the current implementation of pipe and compose (:
````typescript
const cx = (...fns) => fns.reduceRight(compose)
const px = (...fns) => fns.reduce(pipe)
````

Comment (Frikki) [reply packages/prelude/src/function.js]: @nissoh, I think you’re right about the performance loss, unless V8 has become extremely clever.

The current implementation cannot be used the way you describe. Try yourself:

```
const compose = (f, g) => x => f(g(x))

const cx = (...fns) => fns.reduceRight(compose)

const add2 = x => x + 2

const multiply = (x, y) => x * y

cx(add2, multiply)(4, 10) // NaN
```

nissoh: @Frikki yeah, to solve that we have to curry multiply

````typescript
const compose = (f, g) => x => f(g(x))

const cx = (...fns) => fns.reduceRight(compose)
const add2 = x => x + 2
const multiply = curry2((x, y) => x * y)
cx(add2, multiply)(4)(10) // 60
````

TrySound:
@briancavalier Here it is. Flow-bin 55 released. Docs can be waited forever, but [recompose](https://github.com/acdlite/recompose/pull/523) already going to use it (feature was originally requested by recompose maintainer).

briancavalier:
@TrySound Thanks for the link.  I'm guessing `$Compose` supports composing an arbitrary number of functions (not just 2)?  Does it allow the resulting composed function to be of the same arity as the last function passed in?  Maybe we have to wait for docs to appear?

On a related note, the core team recently had a big discussion about `pipe` and composing 2 vs many functions in the context of the intent and goals of `@most/prelude`.  I'll try to summarize here soon, unless someone beats me to it.

Frikki:
@TrySound, to summarize the discussion that @briancavalier mentioned, the core team agreed that _only functions used internally by `@most` should enter `@most/prelude`_. The reasons given are:

1. Avoid requests for new functions, and
2. less maintenance and support for functionality that isn’t core.

Thus, as `pipe` _isn’t_ used internally, we don’t see any compelling reason to include the function. There are already fine libraries out there, e.g., `ramda` and `@typed/prelude`, that are designed specifically for functional programming.

As it stands, we are yet to include a notice in the documentation to avoid new requests and PRs submitted in vain.

TrySound:
Okay. Thanks for answer.

nissoh:
@Frikki, thanks for the update, i think that decision seems reasonable

PR #137, opened 2017-09-22
chore(DOCS): add multicast

davidchase:


davidchase:
woo hooo ship them all 👍

docs/api.rst
```diff
@@ -1064,6 +1064,25 @@ Recover from a stream failure by calling a function to create a new :ref:`Stream
...
+.. _multicast:
+
+multicast
+`````````
+
+.. code-block:: haskell
+
+  multicast :: Stream a -> Stream a
+
+Returns a :ref:`Stream` equivalent to the original, but which can be shared more efficiently among multiple consumers. ::
```
Frikki: Corrrect:
> Returns a :ref:`Stream` equivalent to the original but which can be shared more efficiently among multiple consumers.

We don’t have two independent clauses, so no comma before _but_. Even without _but_, we have an adjective clause that modifies _Stream_; _which_ is a relative pronoun referring to the antecedent _Stream_.

docs/api.rst
```diff
@@ -1064,6 +1064,25 @@ Recover from a stream failure by calling a function to create a new :ref:`Stream
...
+
+  stream:             -a-b-c-d->
+  multicast(stream):  -a-b-c-d->
+
+Using multicast allows you to build up a stream of maps, filters, and other transformations, and then share it efficiently with multiple observers.
```
Frikki: Discard _Using_:
> Multicast allows you to build up a stream of maps, filters, and other transformations, and then share it efficiently with multiple observers.

Frikki:
I’m concerned about the wording "... can be shared more efficiently among multiple consumers." Is it even feasible to share a non-multicast stream with multiple consumers?

Frikki:
I am content now with the current wording. This example by David demonstrates: https://www.webpackbin.com/bins/-KunuEEJxMtd-L48fPBk

PR #139, opened 2017-10-02
feat(examples): Add simple example counter

briancavalier:
I figured I'd get the ball rolling on adding some examples.  This is the usual, trivial inc/dec counter.  It's probably more important to discuss:

1. Is this the right place to put examples?
2. Is the structure of the example reasonable? i.e. it's very self-contained, with it's own package.json, package-lock.json, and rollup.config.js.  Another option may be to share those one level up in the examples folder, as we've done in [mostjs/examples](https://github.com/mostjs/examples).  Might be easier maintenance if we do.

### DIFF
--- a/examples/counter/rollup.config.js
+++ b/examples/counter/rollup.config.js
@@ -1,5 +1,6 @@
 import node from 'rollup-plugin-node-resolve'
 import flow from 'rollup-plugin-flow'
+import pkg from './package.json'
 
 export default {
   input: 'src/index.js',
@@ -8,7 +9,7 @@ export default {
     node()
   ],
   output: {
-    file: 'dist/index.js',
+    file: pkg.main,
     format: 'iife'
   }
 }
--- a/examples/counter/rollup.config.js
+++ b/examples/counter/rollup.config.js
@@ -1,15 +1,11 @@
 import node from 'rollup-plugin-node-resolve'
-import buble from 'rollup-plugin-buble'
 import flow from 'rollup-plugin-flow'
 
 export default {
   input: 'src/index.js',
   plugins: [
     flow(),
-    buble(),
-    node({
-      modulesOnly: true
-    })
+    node()
   ],
   output: {
     file: 'dist/index.js',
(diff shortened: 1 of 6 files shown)


davidchase:
What if we just offer the src/index.js and a link to a webpackbin where the viewer could play with the demo and also download the whole package that would be ready to work via a`yarn/npm install` that we its much less for us to maintain ?

examples/counter/rollup.config.js
```diff
@@ -0,0 +1,18 @@
+import node from 'rollup-plugin-node-resolve'
+import buble from 'rollup-plugin-buble'
+import flow from 'rollup-plugin-flow'
+
+export default {
+  input: 'src/index.js',
+  plugins: [
+    flow(),
+    buble(),
```
TrySound: Why buble for examples? es6 works great in the browsers and node.

examples/counter/src/index.js
```diff
@@ -0,0 +1,20 @@
...
+import { click } from '@most/dom-event'
+
+const fail = s => { throw new Error(s) }
+const qs = (s) => document.querySelector(s) || fail(`${s} not found`)
+
+const incButton = qs('[name=inc]')
+const decButton = qs('[name=dec]')
+const value = qs('.value')
+
+const inc = map(_ => 1, click(incButton))
```
TrySound: I think this is the case for `constant` combinator.

examples/counter/rollup.config.js
briancavalier: Sounds good to me.

briancavalier:
@davidchase Can you say a bit more about exactly how your idea would work?  I'm generally in favor of anything that makes it easier for users to try the examples, as long as it doesn't add (much) maintenance or tooling complexity.

davidchase:
After discussions I think we can for sake of not delaying v1 move with this example but ideally the setup would just require a JavaScript bin that supports flow or TS and we would just have a single EXAMPLES.md file that would link to these bins where developers could explore the examples as well as download the bin to run locally etc maybe something akin to https://hyperapp.glitch.me/ could work as well though I think glitch is still in beta and working out different bugs

Issue #138, opened 2017-10-01
[discussion] Tree shake-ability


Andarist:
At the current state the library, although highly modular (👏), cannot be effectively tree-shaken (well, dead code eliminated to be more exact), because of the heavy `curry` usage. It is like that because bundlers/minifiers cannot know for sure that the call to `curry` is pure and wont cause any side-effects - they cannot be sure that it is safe to remove that call, even if its result stays unused.

UglifyJS has introduced possibility to mark calls as pure with comment annotation - `#__PURE__`. This could be leveraged in the library.

While `#__PURE__` comments are useful today, it's quite ugly to put them into the source code and also it's error-prone, it's easy to forget to put them everywhere during development. I've created a [babel plugin](https://github.com/Andarist/babel-plugin-annotate-pure-calls) which helps with that and automates this task, but at the moment you are using buble and not babel for es2015+ transpilation and Im not sure if you are open to a change and if you want your code to get annotated (in any way, doesnt have to be with a plugin I've created).

So in general - what do you think about this? Would you like me to send a PR with buble -> babel + the plugin?

More reading about the problem - [here](https://iamakulov.com/notes/polished-webpack/)

TrySound:
Also webpack4 will look at `side-effects` in package.json.
https://medium.com/webpack/road-to-webpack-4-week-20-21-1641d03ce06e#dc50
So it's not necessary to annotate code.

Andarist:
I'm aware of the `side-effects` entry for the `package.json`. It's not the only bundler though and the future of `side-effects` in others is uncertain at this point in time - also for considerable time many teams won't be able to migrate to the new webpack's version right after the release.

briancavalier:
Hi @Andarist, thanks for raising this issue!

I'm in favor of anything that helps, as long as we feel the concrete benefit to users outweighs any additional maintenance costs or complexity.  From that perspective, it seems like `side-effects: false` is a no-brainer *once it's released*.

For the PURE annotations, buble satisfies our needs, and we are close to 1.0, so I'd rather not make a major build system change right now.   I'm open to revisiting that after 1.0, though.

Buble does tend to generate leaner code than babel in the first place.  I wonder if the tradeoff of adding babel's overhead, but with improved tree-shaking turns out to yield better results for users *on average*.  Do you have any intuition about that?

Would it be possible to have a rollup version of the plugin?  Or is there some way to use the existing plugin with rollup, without fully committing to babel right now?

Andarist:
[...]

PR #140, opened 2017-10-03
chore(build): proof-of-concept PURE annotations

briancavalier:
Proof of concept for #138, using babel *only* to apply the PURE annotations.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -41,12 +41,14 @@
   "license": "MIT",
   "devDependencies": {
     "@briancavalier/assert": "^3.2.0",
+    "babel-plugin-annotate-pure-calls": "^0.2.0",
     "buba": "^4.0.2",
     "cpy-cli": "^1.0.1",
     "flow-bin": "^0.53.1",
     "mocha": "^3.4.2",
     "nyc": "^11.0.2",
     "rollup": "^0.43.0",
+    "rollup-plugin-babel": "^3.0.2",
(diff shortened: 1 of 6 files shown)


briancavalier:
Ok, I've just spent an hour trying and failing to get a working setup of this that allows:

1. Removing buble and using rollup and babel
1. Building dist with babel es2015 loose mode
1. Running unit tests also with loose mode (so we know we're testing the same thing we're distributing)
1. Adding pure annotations

I can't find any combination of tools and plugins that makes all of the above work.  There seem to be various inconsistencies with es2015, es2015-rollup, babel options in package.json and babel options in rollup configs.  For example, es2015-rollup doesn't support any options, so it seems impossible to enable loose mode when using it.

**If you'd like to help, rather than adding a "but did you try X?" comment, please help by opening a PR with a working setup** 😄

Thanks!

TrySound:
This works in my env. Didn't try with pure annotations. I guess just adding plugin should work.
```js
{
  presets: [
    ["env", {
      modules: false,
      loose: true,
      forceAllTransforms: true
    }]
  ],
  plugins: [
    "external-helpers"
  ]
}
```

Andarist:
I'll send PR addressing list mentioned in [here](https://github.com/mostjs/core/pull/140#issuecomment-334112941) some time later today/tomorrow.

briancavalier:
Closing in favor of #143

PR #141, opened 2017-10-03
chore(build): stop bundling deps into builds

briancavalier:
In the same vein as https://github.com/mostjs/hold/pull/22, this stops bundling deps into each package.  Without this change, for example, we end up with a copy of `@most/prelude` functions in `@most/core`, `@most/disposable`, and `@most/scheduler`.  An application that uses `@most/core` would ultimately end up with 3 copies of some `@most/prelude` functions!  I verified this using the example in #139.  In that tiny example, it reduced the (unminified and non-gzipped) size by 10k.  Perhaps more importantly, it ended up with no duplicate functions.

### DIFF
--- a/packages/core/rollup.config.js
+++ b/packages/core/rollup.config.js
@@ -5,6 +5,11 @@ export default {
   plugins: [
     buble()
   ],
+  external: [
+    '@most/scheduler',
+    '@most/disposable',
+    '@most/prelude'
+  ],
   targets: [
     {
       dest: 'dist/mostCore.js',
--- a/packages/core/rollup.config.js
+++ b/packages/core/rollup.config.js
@@ -1,20 +1,21 @@
 import buble from 'rollup-plugin-buble'
-import nodeResolve from 'rollup-plugin-node-resolve'
 
 export default {
   entry: 'src/index.js',
   plugins: [
-    buble(),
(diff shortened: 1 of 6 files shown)


Frikki:
Excuse my ignorance on the subject, but how does this work? How are the functions used from `@most/prelude` in, e.g., `@most/core` if they "aren't available"?

TrySound:
@Frikki They should be added on a page if you use umd. Or they will be bundled as dependencies with rollup/webpack/browserify automatically without duplicating.

packages/core/rollup.config.js
```diff
@@ -1,20 +1,21 @@
...
+    buble()
   ],
   targets: [
     {
       dest: 'dist/mostCore.js',
       format: 'umd',
       moduleName: 'mostCore',
-      sourceMap: true
+      sourceMap: true,
+      globals: {
```
Andarist: u should also put:
`external: ['@most/scheduler', '@most/disposable', '@most/prelude'],`
at the top level of the `rollup.config.js` to prevent nasty console.warns

briancavalier: Thanks @Andarist.  Added external to relevant rollup configs in f64674a.

PR #142, opened 2017-10-04
feat(withLatestFrom): add pickLatestFrom and zipLatestFrom

briancavalier:
I figured I'd put a stake in the ground and see what happens ... maybe (as a friend of mine says) I'll end up falling on it, lol.  This is definitely POC quality and would need to be cleaned up before merging.  I'm hoping it helps with discussion, though.

This PR removes sample, in favor of adding pickLatestFrom and zipLatestFrom, where zipLatestFrom is basically sample with flipped stream arguments.

"Sample" has a clear connotation for continuous values, but not really for events: when an event is not occurring (which is the vast majority of the time), there is no value to sample.  "Sample" is used often in signal processing as analog -> digital.  There is no analog signal in this case, so, in reality, "sample" doesn't make much sense.  We were basically pretending to treat one of the event streams like an analog value.  I think everyone will be better off if we just treat it like what it is, and put some energy into considering how to integrate "real" analog values, maybe something like [this](https://github.com/briancavalier/most-behave) into the mostjs ecosystem.


packages/core/src/index.js.flow
```diff
@@ -77,10 +77,13 @@ declare export function merge <A> (s1: Stream<A>, s2: Stream<A>): Stream<A>
 declare export function merge <A> (s1: Stream<A>): (s2: Stream<A>) => Stream<A>
 declare export function mergeArray <A> (ss: Array<Stream<A>>): Stream<A>
 
-declare export function sample <A, B, C> (f: (A, B) => C, sampler: Stream<A>, sampled: Stream<B>): Stream<C>
-declare export function sample <A, B, C> (f: (A, B) => C): (sampler: Stream<A>, sampled: Stream<B>) => Stream<C>
-declare export function sample <A, B, C> (f: (A, B) => C, sampler: Stream<A>): (sampled: Stream<B>) => Stream<C>
-declare export function sample <A, B, C> (f: (A, B) => C): (sampler: Stream<A>) => (sampled: Stream<B>) => Stream<C>
+declare export function withLatestFrom <A, B> (values: Stream<A>, stream: Stream<B>): Stream<A>
```
TylorS: I wouldn't bother with the second type parameter.
`withLatestFrom <A> (values: Stream<A>, stream: Stream<any>): Stream<A>`

packages/core/type-definitions/combinator/withLatestFrom.d.ts
```diff
@@ -0,0 +1,13 @@
+import { Stream } from '@most/types';
+
+export function withLatestFrom <A, B> (values: Stream<A>, s: Stream<B>): Stream<A>
```
TylorS: Same as the comment above

packages/core/src/index.js.flow
briancavalier: Yeah, seems reasonable in this case :+1: ... I've just gotten in the habit of avoiding `any`

TrySound: Maybe better `*`?

briancavalier:
Ah, I did flip the parameter order in this PR as we discussed, so I think its (note return type):

```js
// pickLatestFrom :: Stream a -> Stream b -> Stream a
const pickLatestFrom = (streamWhoseLatestValueWillBePicked: Stream<A>, streamWhoseValuesWillBeReplaced: Stream<B>): Stream<A> => ...
```

packages/core/src/index.js.flow
briancavalier: Hmmm, I don't see any advantage to `*` over `any` in this particular case.  The type is completely ignored.  Do you see any advantage _in this particular case_?

TrySound: Just not using `any` :)

PR #143, opened 2017-10-04
Babel MVP

Andarist:
Proof of concept for https://github.com/mostjs/core/pull/140#issuecomment-334112941

Not yet ready to merge!

Extending an `Error` is an issue for babel - had to add for the time being custom transform for this. It's always possible to write buble-like class transform, shouldnt be a problem at all.

It should be easy to compare outputs (before and after) and discuss differences as you guys create "flat bundles" for the distribution


packages/core/.babelrc
```diff
@@ -0,0 +1,5 @@
+{
```
Andarist: in babel7 (soon-ish) it will be possible to use `.babelrc.js` directly and therefore omit those `.babelrc` entirely

packages/disposable/.babelrc.js
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = {
+  presets: [
+    ['env', {
+      modules,
+      loose: true,
+    }]
+  ],
+  plugins: [
+    ['transform-builtin-extend', {
```
Andarist: the custom plugin for extending built-ins - as mentioned above

package.json
```diff
@@ -16,6 +16,26 @@
   "author": "brian@hovercraftstudios.com",
   "license": "MIT",
   "devDependencies": {
-    "lerna": "^2.1.2"
+    "@briancavalier/assert": "^3.2.0",
+    "babel-core": "^6.26.0",
```
Andarist: ive hoisted all common packages to the root `package.json` - according to lerna team this should result in faster bootstraps

package.json
```diff
@@ -16,6 +16,26 @@
...
+    "babel-plugin-annotate-pure-calls": "^0.2.0",
+    "babel-plugin-external-helpers": "^6.22.0",
+    "babel-plugin-transform-builtin-extend": "^1.1.2",
+    "babel-preset-env": "^1.6.0",
+    "babel-preset-power-assert": "^1.0.0",
```
Andarist: couldnt find if it was used before or if this is some leftover from the past, ive kept it for now but it seems this can be either removed or added to the used presets in `prelude` package

packages/core/.babelrc.js
```diff
@@ -0,0 +1,13 @@
+const modules = process.env.NODE_ENV === 'test' ? 'commonjs' : false
+
+module.exports = {
+  presets: [
+    ['env', {
+      modules,
+      loose: true,
```
TrySound: `forceAllTransforms: true`

TrySound:
Better to merge [this](https://github.com/mostjs/core/pull/141) first

Andarist:
Sure thing, I just didn't want to wait before creating a PR, when the other one gets merged in I'm gonna rebase this one.

packages/core/.babelrc.js
Andarist: was thinking about it, but aint sure if there is any benefit in that if we do not provide non-default targets

TrySound: I think it's same target like `uplify` before. At least if defaults will be changed then nothing will be broken.

Andarist: > I think it's same target like uplify before. 

Just a confirmation - that's exactly what it is

> At least if defaults will be changed then nothing will be broken.

Good point, gonna add this later.

TrySound: I meant semantically the same too.

briancavalier:
Thanks for working on this, @Andarist!  Apologies for creating a rebase headache by merging #141.

package.json
briancavalier: Cool.  I was under the impression that each package had to list its own deps and then use `bootstrap --hoist`.  This is much better.

package.json
briancavalier: Oops.  Yes, we can get rid of it.

packages/core/.babelrc.js
briancavalier: Am I understanding [`forceAllTransforms`](https://github.com/babel/babel-preset-env/blob/master/README.md#forcealltransforms) correctly in that it essentially causes `babel-preset-env` to behave like `babel-preset-es2015`?

package.json
Andarist: I believe that each pkg should still lost its own deps, at least i treat lerna packages being in single repository just some kimd of implementation detail. Ideally i always want to keep them separate so the pkg's dir could copied/moved and still keep working without changes.

Also each pkg need to have its deps with executables listed so they can be linked properly and used from npm scripts

Gathering all deps in root package.json is an additional step which according to lerna docs (i think) allows lerna for faster bootstraps but i aint sure why.

packages/core/.babelrc.js
Andarist: Kinda, its job is to output pure es5 which can be ofc understood by subsequent tools (like uglifyjs)

Preset2015 has only es2015 plugins in it, but preset-env has also other yearly presets and also it has (or soon it will, aint sure) shipped proposals plugins (like when browsers implement language features from stage 3 or something).

Andarist:
Dont worry, i knew it will be needed later from the very beginning, also shouldnt be particularly headache-y ;)

package.json
Andarist: https://github.com/lerna/lerna/blob/master/FAQ.md#root-packagejson

> The root can also hold all the "hoisted" packages, which speeds up bootstrapping when using the --hoist flag.

Andarist:
Ok, I've rebased this PR, please keep in mind that this is not merge-ready, because of the usage of custom (non es5 compatible) `babel-plugin-transform-builtin-extend`. If you decide that you want to merge this PR, it should be rewritten first - my proposal is to just create buble-compatible class transform. Ofc I offer to write it 😉

davidchase:
hi @Andarist this is quite interesting setup you have here, is there anyway to compare the results of this setup to our current setup to see what wins we get?

Andarist:
The diff can be seen [here](https://github.com/Andarist/core/pull/1/files?diff=split). Unfortunately it is not easily comparable. I've studied it for some time and mostly (haha 🤣) the difference is in class transform:
- buble generates a little bit simpler classes
- buble handles built-in extension just fine (you extend Error), atm babel doesnt handle this too well
- buble for simple cases doesnt wrap prototype pattern in an IIFE (it does it only when u use inheritance) - this is not a good thing, assignments to the unused function (such as prototype method assignment to the constructor) prevents dead code elimination
- with babel you can use rich ecosystem of plugins (ive added one for adding `#__PURE__` annotations), or write ur own

Ofc as mentioned above - ur original code style is lost in the output when using babel.

davidchase:
any examples of bundle size comparisons ?

Andarist:
Sure - I can check them out and report back later, but diffs will be a little bit off, because a PR is not finished due to the issues described regarding class transform.

I always measure such things by inspecting minified+gziped UMD bundles, is that ok or should I measure in some other way?

davidchase:
I think that would be nice to see you could use https://github.com/mostjs/core/tree/master/examples/counter as a way to compare

briancavalier:
> I always measure such things by inspecting minified+gziped UMD bundles, is that ok

@Andarist Yep, that'd be just fine.  Thank you for all your work on this so far!  Looking forward to seeing size comparisons.

briancavalier:
Ping @Andarist.  We're definitely still interested in seeing a size comparison.  Is there anything else we can do to help?

Andarist:
Nah, not really - I will send comparisons as soon as I can, just couldn't focus at this PR lately as I had some other issues to tackle on.

briancavalier:
Thanks @Andarist.  No worries 😄

Andarist:
All results for UMD minified and gzipped bundles 

./packages/core/dist/mostCore.js
pre - 9.27 kB
post - 9.47 kB

./packages/disposable/dist/index.js
pre - 1.59 kB
post - 1.87 kB

./packages/prelude/dist/index.js
pre - 1.06 kB
post - 1.05 kB

./packages/scheduler/dist/index.js
pre - 3.3 kB
post - 3.31 kB

Difference is there, although quite small and as I've written before - alternative class transform should be written before merging this in, which potentially could decrease the those size differences.

I've also noticed that only the `core` package is building uglified bundle, other do not do that. Is it intended?

briancavalier:
@Andarist Whew!  Now that 1.0 is released, I'm going to spend some time on this.  My plan is to build https://github.com/mostjs/core/tree/master/examples/counter using master and this branch so we can see the size difference in an application.  Granted, it's a _trivial_ application, but I think that's a good place to start since it should result in a fairly large size difference (at least, I think it should!).  If it does then I think we'll feel good about moving ahead.

Andarist:
Recent versions of rollup might be able to tree-shake some `curry` calls so actually `#__PURE__` might help less there.

The experiment should use webpack with ModuleConcatenationPlugin on and uglifyjs, those settings should give better tree-shaking results - webpack alone actually has quite poor algorithm for this stuff.

I will try to rebase the branch a little bit later.

briancavalier:
I made a couple local branches, one based on this PR branch (rebased to latest master), and one based on latest master.  Both add `rollup-plugin-uglify` and `uglify-es` to the examples dir rollup config.  I built the counter app in both branches, and verified that both still work correctly in Chrome, with no errors or warnings in the console.

Here are the results with just uglify, and with uglify+gzip:

```
❯ /bin/ls -l
total 168
-rw-r--r--  1 brian  staff  38834 Nov 28 08:13 index.base.js
-rw-r--r--  1 brian  staff   7792 Nov 28 08:24 index.base.js.gz
-rw-r--r--  1 brian  staff     89 Nov 28 08:18 index.js.map
-rw-r--r--  1 brian  staff  21925 Nov 28 08:18 index.pure.js
-rw-r--r--  1 brian  staff   4581 Nov 28 08:24 index.pure.js.gz
```

`index.base.*` is current master without PURE annotations, and `index.pure.*` is with PURE annotations.  The non-gzip pure version is `21925 / 38834 = 56%` of the size of the corresponding base version.  The gzip pure version is `4581 / 7792 = 58%` the size of the corresponding base version.

The savings are quite significant.  Again, this is for a trivial app that uses a tiny amount of the total API surface area.  Obviously, an app that uses more of the API will see less of a savings ... but that's the whole point: You get exactly what you opt into, and no more.

This seems compelling enough to me to proceed, but I'd love to hear everyone else's thoughts.

briancavalier:
@Andarist Ok, thanks for the info about recent rollup.  I'll try to run the experiment again with the latest version and post more results.  We should do a similar test with webpack.  The savings are probably less, but its worth seeing.

I have to switch to day job mode now, but I'll push my branches when I have a chance in case anyone else wants to experiment.

Andarist:
If you have already rebased the branch - feel free to push-force it here :)

Keep in mind that I've added extra babel transform, so the built-ins like Error can be extended (buble's transform allowed it out of the box) and probably it would be better (?) to create a compatible babel transform for this or something or at least investigate what the added transform actually do and decide if it's worth using.

Also my plugin `babel-plugin-annotate-pure-calls` is somewhat dangerous - it should work for majority of use cases though, but the caution is advised and of course if you find any problem with it I will be happy to fix it.

briancavalier:
> If you have already rebased the branch - feel free to push-force it here

:+1: I hope to do that soon.

> Keep in mind that I've added extra babel transform, so the built-ins like Error can be extended

I think there's [only one spot where we extend a builtin](https://github.com/mostjs/core/blob/master/packages/disposable/src/disposeAll.js#L44).  Maybe we should just find another solution.  For example, we could just use plain prototypal inheritance, or or we could construct an Error and forcibly attach a stack or message that wraps up the stack traces of all the original errors [as this is already doing now](https://github.com/mostjs/core/blob/master/packages/disposable/src/disposeAll.js#L55).  I def don't see this as a blocker, though.

> babel-plugin-annotate-pure-calls is somewhat dangerous

Is there more information about particular situations where it could cause a problem?

Andarist:
> Is there more information about particular situations where it could cause a problem?

At the moment I know only about a single issue - computed properties without loose mode

**input**
```js
const example = {
	['foo']: 1,
  	['bar']: 2,
}
```

**output**
```js
var _example;

function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }

var example = (_example = {}, /* #__PURE__ */ _defineProperty(_example, 'foo', 1), /* #__PURE__ */ _defineProperty(_example, 'bar', 2), _example);
```

Even if `example` gets used, `_defineProperty` calls will get dropped because their direct result is unused and uglifyjs doesnt know that in fact they mutate the object. `loose` mode uses simple assignments for computed properties transform, so everything is good.

I suspect there might be some similar cases, but I haven't yet got any other reports.

briancavalier:
@Andarist I'm realizing that I have no idea how to push my changes to a branch that originates in your fork.  Can you help me learn how to do that?

Andarist:
This is something I usually have to check a little bit, but it should boil down to something like:
```sh
git add remote Andarist git@github.com:Andarist/core.git 
git push --force Andarist babel:babel
```
If it doesn't work I will check later.

briancavalier:
@Andarist Thanks.  I'm not having any luck, though 😢 

(I had already done `git fetch Andarist`)

```
~/Projects/cujojs/@most/core/examples Andarist-babel
❯ git remote -v
Andarist	git@github.com:Andarist/core.git (fetch)
Andarist	git@github.com:Andarist/core.git (push)
origin	git@github.com:mostjs/core.git (fetch)
origin	git@github.com:mostjs/core.git (push)

~/Projects/cujojs/@most/core/examples Andarist-babel
❯ git push Andarist babel:babel
error: src refspec babel does not match any.
error: failed to push some refs to 'git@github.com:Andarist/core.git'
```

I didn't use `-f` so that I could see if it was right before force pushing.  Is there another way I can get the changes to you?  For example, maybe I can PR back to your branch?

briancavalier:
I pushed my rebased branch [here](https://github.com/mostjs/core/tree/Andarist-babel) for now.

Andarist:
PR would be an option for more basic use case, but this is a rebased branch and I'm not brave enough to merge  2 alternate git histories 😂 

Thanks for the link to the updated branch, gonna play with it in a minute :)

Andarist:
If your branch is named `Andarist-babel`, then try this:
```
git push Andarist Andarist-babel:babel
```

briancavalier:
I didn't get a chance to try that yesterday.  I'll try it now.  🤞

briancavalier:
:+1: I ran it without `-f` just to try, and it seems to work.  Going to run it _with_ `-f` now!

briancavalier:
And there we go.

Next, I'm going to try to get rid of the need for extending Error.  I'll do that in another PR so we can all see what it looks like before making a decision about it.

Andarist:
👍 that would be a breaking change though, wouldnt it?

briancavalier:
I think it depends on exactly what we do.  The particular error thrown by [`disposeAll` isn't part of the public API](http://mostcore.readthedocs.io/en/latest/api.html#most-disposable), so I feel it's ok to change it.  Using native prototypal inheritance instead of `extends` may be a solution as well.

Andarist:
Then both options seems fine to me, certainly better to use prototypal inheritance for a single use case than adding an additional (somewhat) complex plugin. If you follow that path it's advised to use IIFE to output a class into a single variable.

briancavalier:
For comparison:

1. 85031a7faf7023ec0484cb3dd7396338f6bb9089 uses prototypal inheritance.  This seems pretty good to me.
1. 7bcb3789c1b670e8f73f3241dea75510afaf12a9 uses a regular `Error`.  This is slightly a "bigger" change.  I don't see much advantage, but it also seems fine to me.

Andarist:
Personally I like 1st better, seems like a cleaner change and being more straightforward about its purpose.

I'd only change it to:
```js
export const DisposeAllError = (Error => {
  function DisposeAllError(message, errors) {
    Error.call(this, message)
    this.message = message
    this.name = DisposeAllError.name
    this.errors = errors

    if (Error.captureStackTrace) {
      Error.captureStackTrace(this, DisposeAllError)
    }

    this.stack = `${this.stack}${formatErrorStacks(this.errors)}`
  }	

  DisposeAllError.prototype = Object.create(Error.prototype)

  return DisposeAllError
})(Error) 
```

briancavalier:
@Andarist Agreed about the first option.

What's the reason for the IIFE in this particular case?  It's not clear to me what it's defending against.

Andarist:
Having this in top level scope:
```js
DisposeAllError.prototype = Object.create(Error.prototype)
```
will prevent tree-shaking of `DisposeAllError` in most cases. Because assignments like this are treated as effectful, especially considering the fact that `Object.create` may not be whitelisted as "pure" in dead code eliminators like uglifyJS (it might be, aint sure about this, would have to check).

briancavalier:
> will prevent tree-shaking of DisposeAllError in most cases

Ah, ok, thanks for explaining.  I'll add the IIFE.

briancavalier:
Ok, there we go: https://github.com/mostjs/core/commit/1019bda39bf6780d7c5b97dbf0fc64f9824355fb

Do you want me to cherry pick that onto this branch, and remove the babel plugin?

Andarist:
Yeah 👍  that would simplify working with this branch :)

briancavalier:
Done! Cherry-picked prototypal inheritance and removed builtin-extend plugin on this branch.

Andarist:
Cool! This PR is ready to merge then, if you want to proceed with it. Let me know if it needs any other adjustments.

briancavalier:
Here are results with webpack:

```
❯ /bin/ls -lrt counter/dist/
total 184
-rw-r--r--  1 brian  staff  25121 Dec  7 08:42 index.webpack.pure.js
-rw-r--r--  1 brian  staff   5481 Dec  7 08:44 index.webpack.pure.js.gz
-rw-r--r--  1 brian  staff  42553 Dec  7 08:55 index.webpack.base.js
-rw-r--r--  1 brian  staff   9486 Dec  7 08:56 index.webpack.base.js.gz
```

25121 / 42553 = ~59% of original non-gzip
5481 / 9486 = ~58% of original gzip

And here's the webpack config I used:

```js
const path = require('path')
const webpack = require('webpack')
const UglifyJSPlugin = require('uglifyjs-webpack-plugin')

module.exports = {
  entry: './counter/src/index.js',
  output: {
    filename: 'index.js',
    path: path.resolve(__dirname, 'counter/dist')
  },
  resolve: {
    mainFields: [ 'module', 'jsnext:main', 'browser', 'main' ]
  },
  devtool: 'source-map',
  plugins: [
    new webpack.optimize.ModuleConcatenationPlugin(),
    new UglifyJSPlugin()
  ]
}
```

briancavalier:
I pushed the two branches I used for the tests.  Be sure to `npm install` in _both_ the top level dir _and_ the examples dir.  You may also want to verify that for the PURE test, you have indeed ended up with PURE annotations in `examples/node_modules/@most/core/dist/index.es.js`.  When I did the test the first time, somehow I missed a step and ended up without them--thus making it appear as if there was no savings at all.

test without PURE: https://github.com/mostjs/core/tree/x-webpack-tree-shake-master
test with PURE: https://github.com/mostjs/core/tree/x-webpack-tree-shake-pure

Andarist:
Cool comparisons! Would be good to test webpack without `ModuleConcatenationPlugin` too, I predict though that it won't give even nearly similar savings - should save a little bit of something though. Good news is that webpack@4 is gonna have `ModuleConcatenationPlugin` enabled **by default** for production builds.

briancavalier:
Without ModuleConcatenationPlugin (See `index.webpack.pure-no-concat.*`), surprisingly similar results.  It may be because this is a trivial application, though.

```
❯ /bin/ls -lr counter/dist/
total 256
-rw-r--r--  1 brian  staff   5481 Dec  7 08:44 index.webpack.pure.js.gz
-rw-r--r--  1 brian  staff  25121 Dec  7 08:42 index.webpack.pure.js
-rw-r--r--  1 brian  staff   5655 Dec  7 12:05 index.webpack.pure-no-concat.js.gz
-rw-r--r--  1 brian  staff  26122 Dec  7 12:04 index.webpack.pure-no-concat.js
-rw-r--r--  1 brian  staff   9486 Dec  7 08:56 index.webpack.base.js.gz
-rw-r--r--  1 brian  staff  42553 Dec  7 08:55 index.webpack.base.js
```

webpack config:

```js
const path = require('path')
// const webpack = require('webpack')
const UglifyJSPlugin = require('uglifyjs-webpack-plugin')

module.exports = {
  entry: './counter/src/index.js',
  output: {
    filename: 'index.js',
    path: path.resolve(__dirname, 'counter/dist')
  },
  resolve: {
    mainFields: [ 'module', 'jsnext:main', 'browser', 'main' ]
  },
  devtool: 'source-map',
  plugins: [
    // new webpack.optimize.ModuleConcatenationPlugin(),
    new UglifyJSPlugin()
  ]
}
```

briancavalier:
@Andarist Huge thanks for starting the conversation in #138 and for this PR. @TrySound, thank you as well for the help and discussion.

I'm having a real "this is why open source is awesome" moment right now about the awesome collaboration 😄 ... and I'm excited that we'll be able to release such a concrete improvement for `@most/core` users!

I'm gonna go through the diff one last time to make sure we've tied up all loose ends.

Andarist:
Actually results are similar because you are producing a flat bundle 👏 

If you'd transpile file by file and produce a "mirror" directory structure results would be a lot worse probably and that would be exposed even more by inspecting a "trivial app" because in such a trivial app tree shaking should do more than for more complex app which would probably use a lot more of the core and thus benefits would be smaller.

> I'm having a real "this is why open source is awesome" moment right now about the awesome collaboration 😄 ... and I'm excited that we'll be able to release such a concrete improvement for @most/core users!

Thanks for kind words :) I have just explored this area quite a bit lately and observed what prevents and what not tree shaking algos, so I've looked on several libs and tried to improve the situation where I could

briancavalier:
Just published `@most/core` 1.1.0 with PURE annotations 🚀

PR #144, opened 2017-10-05
docs(concepts): Add finite, infinite, failed stream info, periodic docs

briancavalier:
Add some info on finite and infinite streams, and attempt to clarify stream failures vs. application errors.  Docs for `periodic` were missing, so I added them along the way.

I also fixed some sphinx errors.  It seems to get confused when you have a `:ref:` or inline code/preformatted span that *isn't immediately surrounded by spaces*.  For example, these are problematic:

```rst
Some :ref:`Stream`s blah blah

Some ``Stream``s blah blah
```

I don't have a good solution to the second one.  The first one can be solved by using the explicit text+reference syntax:

```rst
Some :ref:`Streams <Stream>`
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -268,7 +268,7 @@ periodic
 
 Create an infinite :ref:`Stream` containing events that occur at a specified period. ::
 
-  periodic(3): .--.--.--.-->
+  periodic(3): x--x--x--x-->
 
 The first event occurs at time 0, and the event values are ``undefined``.
 
@@ -281,7 +281,7 @@ throwError
 
   throwError :: Error -> Stream void
 
-Create a :ref:`Stream` that fails at time 0 with the provided ``Error``.  This can be useful for functions that need to return a :ref:`Stream` and also need to propagate an error. ::
+Create a :ref:`Stream` that fails with the provided ``Error`` at time 0.  This can be useful for functions that need to return a :ref:`Stream` and also need to propagate an error. ::
 
   throwError(X): X
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -36,11 +36,11 @@ Stream
     run :: Sink a -> Scheduler -> Disposable
   }
 
-A ``Stream`` represents a view of events over time. Its ``run`` method arranges events to be propagated to the provided :ref:`Sink` in the future. Each ``Stream`` has a local clock, defined by the provided :ref:`Scheduler`, which has methods for knowing the current time and scheduling future :ref:`Task`s.
+A ``Stream`` represents a view of events over time. Its ``run`` method arranges events to be propagated to the provided :ref:`Sink` in the future. Each ``Stream`` has a local clock, defined by the provided :ref:`Scheduler`, which has methods for knowing the current time and scheduling future :ref:`Tasks <Task>`.
 
-Some ``Stream``s, like :ref:`now`, are simple, while others, like :ref:`combine`, do sophisticated things such as combining multiple ``Stream``s or dealing with higher-order ``Stream``s.
+A ``Stream`` may be simple, like :ref:`now`, or may do sophisticated things such as :ref:`combining <combine>` multiple ``Stream`` s or deal with higher-order ``Stream`` s.
 
-Some ``Stream``s act as event producers, such as from DOM events. A producer ``Stream`` must never produce an event in the same call stack as its ``run`` method is called. It must begin producing items asynchronously. In some cases, this comes for free, such as DOM events. In other cases, it must be done explicitly using the provided ref:`Scheduler` to schedule asynchronous ref:`Task`s.
+A ``Stream`` may act as an event producer, such as a ``Stream`` that produces DOM events. A producer ``Stream`` must never produce an event in the same call stack as its ``run`` method is called. It must begin producing items asynchronously. In some cases, this comes for free, such as DOM events. In other cases, it must be done explicitly using the provided ref:`Scheduler` to schedule asynchronous :ref:`Tasks <Task>`.
 
 .. _Sink:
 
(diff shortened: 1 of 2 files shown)


docs/api.rst
```diff
@@ -255,6 +257,21 @@ Create a :ref:`Stream` containing a single event at a specific time. ::
...
+periodic
+````````
+
+.. code-block:: haskell
+
+  periodic :: Period -> Stream void
+
+Create an infinite :ref:`Stream` containing events that occur at a specified period. ::
+
+  periodic(3): .--.--.--.-->
```
briancavalier: Is the `.` here confusing?  Maybe we should just use `x`?

Frikki: yes and yes

docs/api.rst
```diff
@@ -36,11 +36,11 @@ Stream
     run :: Sink a -> Scheduler -> Disposable
   }
 
-A ``Stream`` represents a view of events over time. Its ``run`` method arranges events to be propagated to the provided :ref:`Sink` in the future. Each ``Stream`` has a local clock, defined by the provided :ref:`Scheduler`, which has methods for knowing the current time and scheduling future :ref:`Task`s.
+A ``Stream`` represents a view of events over time. Its ``run`` method arranges events to be propagated to the provided :ref:`Sink` in the future. Each ``Stream`` has a local clock, defined by the provided :ref:`Scheduler`, which has methods for knowing the current time and scheduling future :ref:`Tasks <Task>`.
 
-Some ``Stream``s, like :ref:`now`, are simple, while others, like :ref:`combine`, do sophisticated things such as combining multiple ``Stream``s or dealing with higher-order ``Stream``s.
+A ``Stream`` may be simple, like :ref:`now`, or may do sophisticated things such as :ref:`combining <combine>` multiple ``Stream`` s or deal with higher-order ``Stream`` s.
```
Frikki: Why the space after \`\`Stream\`\` and before s?

docs/api.rst
```diff
@@ -264,9 +281,7 @@ throwError
 
   throwError :: Error -> Stream void
 
-Create a :ref:`Stream` that fails at time 0 with the provided ``Error``.
-
-This can be useful for functions that need to return a :ref:`Stream` and also need to propagate an error. ::
+Create a :ref:`Stream` that fails at time 0 with the provided ``Error``.  This can be useful for functions that need to return a :ref:`Stream` and also need to propagate an error. ::
```
Frikki: I would switch the order, because it's the `Stream` that fails with the `Error`:

> Create a :ref:`Stream` that fails with the provided \`\`Error\`\` at time 0.

docs/concepts.rst
```diff
@@ -15,12 +15,52 @@ An *event stream* is a time-ordered series of events.  For example, all the mous
...
+^^^^^^^^^^^^^^^^^^^^
+
+An event stream may be finite.  A finite event stream will produce an *end signal* to indicate that it will never produce another event.  When an event stream ends, it will free underlying resources.
+
+.. _Failed Event Streams:
+
+Failed Event Streams
+^^^^^^^^^^^^^^^^^^^^
+
+An event stream may fail.  For example, when an event stream represents a resource, such as a websocket, and the resource fails or closes unexpectedly, the event stream *cannot* produce more events.  When an event stream fails, it will produce a *error signal* to indicate that it cannot produce more events.  The error signal includes a *reason* (an ``Error`` object) describing the failure.
```
Frikki: Correction:
> ... it will produce an *error signal* ...

docs/concepts.rst
```diff
@@ -15,12 +15,52 @@ An *event stream* is a time-ordered series of events.  For example, all the mous
...
+A failed event stream will attempt to free any underlying resources.
+
+The :ref:`recoverWith` operation allows you to handle an event stream failure.
+
+.. _Application Errors:
+
+Stream Failure vs. Application Errors
+`````````````````````````````````````
+
+Stream failures are different from *application errors*.  A stream failure indicates that an event stream *cannot* produce more events.  Application errors may or may not indicate an event stream failure.  For example, an event stream representing messages arriving on a websocket *fails* if the websocket is disconnected unexpectedly due to a wifi drop.  In contrast, an application error may occur in application logic that receives a websocket message and can't process because the message application state has changed.
```
Frikki: Change instances of "websocket" to WebSocket.

Maybe:
> In contrast, an application error may occur in the application logic, when a nonprocessible WebSocket message is received, because the message application state has changed.

docs/concepts.rst
```diff
@@ -15,12 +15,52 @@ An *event stream* is a time-ordered series of events.  For example, all the mous
...
+
+Application error handling is outside the scope of these docs, as it is applications-specific.  However, there are some general strategies for dealing with application errors with event streams:
```
Frikki: Correct:
> ... as it is application specific.

docs/concepts.rst
```diff
@@ -15,12 +15,52 @@ An *event stream* is a time-ordered series of events.  For example, all the mous
...
+
+* Use try/catch, `Promise catch() <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch>`_ to handle the application error and transform it into:
```
Frikki: What does the underscore (`_`) after the link do? Also present below.

docs/api.rst
briancavalier: Sphinx doesn't render it property if you omit the space.  The same is true for `:ref:`s.  The solution to `:ref:`s is to use the expanded form, but I haven't found a solution to the preformatted case.  The only option may be to reword the sentence to use singular instead of plural.

docs/concepts.rst
briancavalier: The underscore is a necessary part of creating a link in sphinx.  See the [sphinx docs for links](http://www.sphinx-doc.org/en/stable/rest.html#hyperlinks).

docs/api.rst
Frikki: _Sigh!_ So be it then.

Issue #147, opened 2017-10-07
Bug: (with/zip)ArrayValues does not stop emitting when expected


TylorS:
Heres a reproducible case, there are additional comments in the code to define the problem.
https://www.webpackbin.com/bins/-KvrqudMj6C-0H494Itg

PR #149, opened 2017-10-09
fix: simplify ZipArrayValuesSink logic

TylorS:
Use take(array.length) to remove duplicate logic in ZipArrayValuesSink

Fix: #147

### DIFF
--- a/packages/core/src/combinator/zipItems.js
+++ b/packages/core/src/combinator/zipItems.js
@@ -1,16 +1,14 @@
 /** @license MIT License (c) copyright 2017 original author or authors */
 
-import { empty } from '../source/empty'
-import { take } from './slice'
 import Pipe from '../sink/Pipe'
+import { isCanonicalEmpty } from '../source/empty'
+import { take } from './slice'
 
 export const withItems = (items, stream) =>
   zipItems(keepLeft, items, stream)
 
 export const zipItems = (f, items, stream) =>
-  items.length === 0 || stream === empty()
-    ? empty()
-    : new ZipItems(f, items, take(items.length, stream))
+  isCanonicalEmpty(stream) ? stream : new ZipItems(f, items, take(items.length, stream))
 
 const keepLeft = (a, _) => a
 
--- a/packages/core/src/combinator/zipItems.js
+++ b/packages/core/src/combinator/zipItems.js
@@ -1,14 +1,16 @@
 /** @license MIT License (c) copyright 2017 original author or authors */
 
 import Pipe from '../sink/Pipe'
-import { isCanonicalEmpty } from '../source/empty'
+import { empty, isCanonicalEmpty } from '../source/empty'
 import { take } from './slice'
 
 export const withItems = (items, stream) =>
   zipItems(keepLeft, items, stream)
 
 export const zipItems = (f, items, stream) =>
-  isCanonicalEmpty(stream) ? stream : take(items.length, new ZipItems(f, items, stream))
+  isCanonicalEmpty(stream) || items.length === 0
+    ? empty()
+    : new ZipItems(f, items, take(items.length, stream))
 
 const keepLeft = (a, _) => a
 
--- a/packages/core/src/combinator/zipItems.js
+++ b/packages/core/src/combinator/zipItems.js
@@ -1,6 +1,7 @@
 /** @license MIT License (c) copyright 2017 original author or authors */
 
 import { empty } from '../source/empty'
+import { take } from './slice'
 import Pipe from '../sink/Pipe'
 
 export const withItems = (items, stream) =>
@@ -9,7 +10,7 @@ export const withItems = (items, stream) =>
 export const zipItems = (f, items, stream) =>
   items.length === 0 || stream === empty()
     ? empty()
-    : new ZipItems(f, items, stream)
+    : new ZipItems(f, items, take(items.length, stream))
 
 const keepLeft = (a, _) => a
 
@@ -36,10 +37,6 @@ class ZipItemsSink extends Pipe {
   event (t, b) {
     const f = this.f
     this.sink.event(t, f(this.items[this.index], b))
-
     this.index += 1
-    if (this.index >= this.items.length) {
-      this.sink.end(t)
-    }
   }
 }
--- a/packages/core/src/combinator/zipItems.js
+++ b/packages/core/src/combinator/zipItems.js
@@ -8,7 +8,7 @@ export const withItems = (items, stream) =>
   zipItems(keepLeft, items, stream)
 
 export const zipItems = (f, items, stream) =>
-  isCanonicalEmpty(stream) ? stream : new ZipItems(f, items, take(items.length, stream))
+  isCanonicalEmpty(stream) ? stream : take(items.length, new ZipItems(f, items, stream))
 
 const keepLeft = (a, _) => a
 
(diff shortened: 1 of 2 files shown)


packages/core/src/combinator/withArrayValues.js
```diff
@@ -1,15 +1,16 @@
...
+import { take } from './slice'
 
 export const withArrayValues = (array, stream) =>
   zipArrayValues(keepLeft, array, stream)
 
 export const zipArrayValues = (f, array, stream) =>
   array.length === 0 || stream === empty()
     ? empty()
-    : new ZipArrayValues(f, array, stream)
+    : new ZipArrayValues(f, array, take(array.length, stream))
```
briancavalier: What do you think about adding a test to cover this, i.e., based on the code where you discovered the bug?

briancavalier: Did you have any thoughts about the wrapping order of take vs. zipArrayValues?  I'm not sure it matters much, but figured I'd ask.

One thing that could be interesting is to fold `empty()` handling into slice()--might be something we want to do regardless of what we decide here.

If we did that, then anything wrapped in slice() (or take, or skip) would automatically gain the empty() or zero-length slice optimization.  I don't think we could fully take advantage of it in this particular case, though, since `take(array.length, new ZipArrayValues(...))` would never trigger a `=== empty()` check inside `take`.  However, it *would* trigger a zero-length slice check--i.e. we could remove the zero-length check in zipArrayValues.

briancavalier:
❤️ the simplification

packages/core/src/combinator/withArrayValues.js
```diff
@@ -35,11 +36,6 @@ class ZipArrayValuesSink extends Pipe {
 
   event (t, b) {
     const f = this.f
-    this.sink.event(t, f(this.values[this.index], b))
-
-    this.index += 1
-    if (this.index >= this.values.length) {
-      this.sink.end(t)
-    }
+    this.sink.event(t, f(this.values[this.index++], b))
```
Frikki: While we all enjoy one-liners, incrementing `this.index` in the array makes it less clear and is easily overlooked, why I advise having the incrementor on a separate line. Also, there could arise issues of the increment operators position—the off-by-one error—i.e., is the index incremented before or after retrieving the value? (Don't worry, I know).

Consider these:
```
x = a+++b -> x = (a++)+b -> x = a + b; a++ // The tokenizer is greedy.
```

The most important rational, however, is that you are actually causing a side effect besides returning a value, and that is what makes the code harder to reason about.

packages/core/src/combinator/withArrayValues.js
TylorS: Yeah, I'm cool with `take(0)` returning `empty()`

packages/core/src/combinator/withArrayValues.js
briancavalier: Agreed.  I vote for `+= 1` on the next line.

packages/core/src/combinator/withArrayValues.js
briancavalier: See #150 for a more thinking on it.

Frikki: I can't believe I missed that :flushed:

briancavalier:
@TylorS Heads-up that I rebased and force pushed this branch.

briancavalier:
73d77c9 adds a regression test based on @TylorS's [webpackbin](https://www.webpackbin.com/bins/-KvrqudMj6C-0H494Itg).

packages/core/src/combinator/zipItems.js
```diff
@@ -9,7 +10,7 @@ export const withItems = (items, stream) =>
 export const zipItems = (f, items, stream) =>
   items.length === 0 || stream === empty()
```
briancavalier: If we merge #151 first, then we can use `isCanonicalEmpty` here.

davidchase: 💯

Frikki: ... which can be done now.

TylorS:
@briancavalier this was rebased from master and force pushed

packages/core/src/combinator/zipItems.js
```diff
@@ -1,16 +1,14 @@
...
+import { take } from './slice'
 
 export const withItems = (items, stream) =>
   zipItems(keepLeft, items, stream)
 
 export const zipItems = (f, items, stream) =>
-  items.length === 0 || stream === empty()
-    ? empty()
-    : new ZipItems(f, items, take(items.length, stream))
+  isCanonicalEmpty(stream) ? stream : new ZipItems(f, items, take(items.length, stream))
```
briancavalier: I think we're missing an opportunity to propagate empty() in the case where items.length === 0.  In that case, rather than returning empty(), this will still return a ZipItems stream.  That ZipItems will _contain_ an empty(), but will prevent any further empty() propagation.

I see two ways to handle it.  First, re-add the other check and explicitly return empty():

```js
isCanonicalEmpty(stream) || items.length === 0 ? empty() : new ZipItems(f, items, take(items.length, stream))
```

Second, flip take and ZipItems.  This reads nicely, but is slightly more wasteful, because it constructs the ZipItems only to throw it away:

```js
isCanonicalEmpty(stream) ? stream : take(items.length, new ZipItems(f, items, stream))
```

In the grand scheme, unless you're calling zipItems in a tight loop with an empty array, you'll never notice the extra ZipItems construction.

Thoughts?

Frikki: > This reads nicely, but is slightly more wasteful ...

What kind of waste are we talking here compared to returning a `ZipItems` instance?

davidchase: i like the 1st ugly one

TylorS: Yeah, I see this a good solution
```typescript
isCanonicalEmpty(stream) ? stream : take(items.length, new ZipItems(f, items, stream))
```

Frikki: The first accesses the `Array.prototype.length` property twice.

TylorS: Updated with the second

briancavalier: > What kind of waste are we talking here compared to returning a ZipItems instance?

I just meant that in the second version, if items.length === 0, it constructs a new ZipItems() and passes it to `take`, which take will then discard and return `empty()`.  So, the waste is the cost of allocation the unnecessary ZipItems().  A memory allocation like that is typically "more expensive" (in memory and cpu) than the second `items.length` access.  But in the grand scheme, the difference is probably tiny.  I doubt it will matter and readability seems more important.

Here's another possible approach that avoids both the double items.length access and the unnecessary ZipItems allocation:

```js
export const zipItems = (f, items, stream) =>
  privateZipItems(f, items, take(items.length, stream))

const privateZipItems = (f, items, stream) =>
  isCanonicalEmpty(stream) ? stream : new ZipItems(f, items, stream)
```

That relies on `take` to return empty() in the cases where items.length === 0 *or* stream is empty().  IOW, it relies on `take` to fold the two cases into one, so that `privateZipItems` only needs to check `isCanonicalEmpty`.

How does that one compare?

Frikki: I like that :heart: Just, I'd call it `internalZipItems` instead of _private_.

briancavalier: Yeah, I'm good with option 1 or 3.

Frikki: It's no. 1 for me.

Issue #150, opened 2017-10-09
Improve empty() annihilation


briancavalier:
There quite a few operations where if at least one input stream is `empty()`, the operation will always return a stream containing zero events.  In such cases, some operations return the canonical `empty()`, and some don't.

### Return empty() when possible

- [x] Tracked in #152

It might be interesting to always return the canonical `empty()` in those cases.  That would allow `empty()` to annihilate other operations more effectively.  It trades some up-front cost for reduction in cpu and memory later when a stream is observed.

```js
const s0 = empty()

// s1 contains no events.  All of these layers could be annihilated by empty()
const s1 = combine(combiner, anotherStream, map(f, filter(p, s0)))

// This isn't true today, but I think it's technically possible
s1 === s0
```

### Make slice fusion smarter about empty()

- [x] Tracked in #151

Somewhat related is that slice, take, and skip could produce `empty()` even without having received it as an input.  Currently, `slice` does [minimal checking](https://github.com/mostjs/core/blob/master/packages/core/src/combinator/slice.js#L34) for a zero-length slice.  However, slice fusion [_doesn't_ check](https://github.com/mostjs/core/blob/master/packages/core/src/combinator/slice.js#L44-L48) to see if it has ended up with a zero-length slice after fusing.  We should refactor it to do that.

### Downsides

The downside, at least right now, is that it requires putting checks in lots of operations (we already check in some operations, but not all).  This seems like something that we'd ultimately like a compiler to be able to do for us.

### Notes

Some notable exceptions:

* merge() and mergeArray() - it's interesting that empty() streams could effectively be dropped.  If mergeArray ends up with a zero length array after dropping empty() streams, then it can also return empty()
* continueWith() - `continueWith(f, empty)` is effectively the same as _calling_ f.  However, it would probably be surprising to users who happen to write a side-effecting `f` that it's called immediately.  Note that recoverWith() _can be_ annihilated by empty()
* scan() - empty() can't annihilate scan entirely, but rather `scan(f, x, empty())` is `now(x)`
* startWith - similar to scan, `startWith(x, empty())` is `now(x)`

briancavalier:
@Frikki :+1: Totally agree.  To that end, see #151 😄

briancavalier:
@Frikki I just created #152 and updated this issue's description with todos for both #151 and #152.

briancavalier:
Closed by #151 and #152

Thanks everyone for the help on this!

PR #151, opened 2017-10-09
feat(slice): improve empty slice detection

briancavalier:
See #150, specifically the "Make slice fusion smarter about empty()" section.

This PR improves detection of empty slices.  If slice detects that it has reached the point where it could only return an empty stream, it will return canonical `empty()`.  Specifically, it now detects these cases _recursively_:

* when it computes end <= start
* when it encounters canonical `empty()`


Frikki:
Closed with PR #151

PR #153, opened 2017-10-13
docs(concepts): Fix grammar in application errors section

briancavalier:


docs/concepts.rst
```diff
@@ -43,11 +43,11 @@ The :ref:`recoverWith` operation allows you to handle an event stream failure.
 Stream Failure vs. Application Errors
 `````````````````````````````````````
 
-Stream failures are different from *application errors*.  A stream failure indicates that an event stream *cannot* produce more events.  Application errors may or may not indicate an event stream failure.  For example, an event stream representing messages arriving on a WebSocket *fails* if the WebSocket is disconnected unexpectedly due to a wifi drop.  In contrast, an application error may occur in application logic that receives a WebSocket message and can't process because the message application state has changed.
+Stream failures are different from *application errors*.  A stream failure indicates that an event stream *cannot* produce more events.  Application errors may or may not indicate an event stream failure.  For example, an event stream representing messages arriving on a WebSocket *fails* if the WebSocket is disconnected unexpectedly due to a wifi drop.  In contrast, an application error may occur in application logic when it receives a WebSocket message that it can't process because the application state has changed.
```
briancavalier: For some reason, GitHub isn't highlighting the diff here very well.  The last sentence is where the change is:

Old:
```
In contrast, an application error may occur in application logic that receives a WebSocket message and can't process because the message application state has changed.
```

New:
```
In contrast, an application error may occur in application logic when it receives a WebSocket message that it can't process because the application state has changed.
```

Frikki: Long lines.

Frikki: Add a `the` before `application logic`:

> ... may occur in the application logic ...

Similar to how you wouldn't say: "An accident may happen in car seat."

briancavalier: Good catch, thanks.  The definite article makes sense since it's referring back to the specific application logic mentioned in the previous sentence, rather than referring to application logic in general.

I split and refactored this paragraph a good bit more.  Let me know what you think.

docs/concepts.rst
```diff
@@ -45,7 +45,9 @@ Stream Failure vs. Application Errors
 
 Stream failures are different from *application errors*.  A stream failure indicates that an event stream *cannot* produce more events.  An application error may or may not indicate an event stream failure.
 
-For example, an event stream representing messages arriving on a WebSocket *fails* if the WebSocket is disconnected unexpectedly due to a wifi drop.  In contrast, an application error may occur in the application logic when it receives a WebSocket message that it can't process because the application state has changed.
+For example, an event stream representing messages arriving on a WebSocket *fails* if the WebSocket is disconnected unexpectedly due to a dropped network connection.
```
Frikki: Correct:
> For example, an event stream, representing messages, arriving on a WebSocket *fails* if the WebSocket is disconnected unexpectedly because of a dropped network connection.

"... representing messages ..." is a parenthetical phrase. It adds information that isn't vitally important.
Because of vs. due to: _Due to_ is always used after a form of the verb _to be_. _Because of_ is not used after a form of the verb _to be_.

docs/concepts.rst
```diff
@@ -45,7 +45,9 @@ Stream Failure vs. Application Errors
...
+
+In contrast, an application error may occur in the application logic when it receives a WebSocket message that it can't process because the application state has changed.  The application must deal with this particular state conflict in an application-specific way, but the WebSocket event stream has not failed.
```
Frikki: > In contrast, an *application error* may occur in the application logic when it receives a WebSocket message that it can't process because the application state has changed.  The application must deal with this particular state conflict in an application-specific way, but the WebSocket event stream has *not* failed.

I think it's good to add the emphases.

docs/concepts.rst
briancavalier: Hmm, making it parenthetical reads a bit awkwardly to me.  What if we rephrase to something like:

"For example, an event stream of WebSocket messages _fails_ if ..."

> Because of vs. due to

:+1: Thanks

briancavalier: Or maybe:

"For example, an event stream of messages from a WebSocket fails if ..."

docs/concepts.rst
briancavalier: Added in 6a31463

docs/concepts.rst
Frikki: Yes. The latter is okay.

briancavalier: Updated in 58ac0b4

Issue #155, opened 2017-10-17
Start a FAQ


briancavalier:
An initial question we could answer:

* What is the relationship between `@most/core` and `most`?

Let's brainstorm a few others here.

davidchase:
- Comparison with other libs ?
- How to use with Vdom libs like React
- How to use with Node callbacks

Frikki:
* How do I produce and consume events from streams? (hint: synchronization)

briancavalier:
Ok, I'm going to call this done, because we did _start_ a FAQ 😄 .  Thoughts on a good place to keep these questions that have been brainstormed here until we have a chance to write answers?

PR #157, opened 2017-10-18
docs(FAQ): start FAQ

briancavalier:
See #155

I figured I'd get the ball rolling.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,0 +1,10 @@
+FAQ
+===
+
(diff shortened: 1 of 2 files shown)


Frikki:
Is it okay if I pull and push to this branch, e.g., to correct grammar, instead of commenting?

briancavalier:
Please do!

briancavalier:
FWIW, I think this single question and answer would be enough to unblock 1.0 ... at least we would the FAQ would exist, and give us a well-defined place to put questions and answers.

PR #161, opened 2017-10-24
Reword Source and sink chains

### LINKED ISSUE #160: Remove "source" as a first class concept in the docs.
There is no first class Source concept in most/core, so the docs shouldn't talk about "source" as if it is first class.

I don't think its harmful to use the word "source" _occasionally_ to refer the the origin of a Stream's events.  For example, the source of the event produced by a `fromPromise` Stream is the promise that was passed in.

[This section](http://mostcore.readthedocs.io/en/latest/concepts.html#source-and-sink-chains) puts too much emphases on "source".  As per slack discussion, we should rename that section to "Streams and Sinks" and talk about Streams instead of "source chains".

Frikki:
Close #160

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -64,7 +64,7 @@ Streams and Sinks
 
 Applying an operation to a stream derives a *new stream* with *new events*. There is *no alteration* to the original stream (*the origin*). Multiple operations compose a *chain of streams*. When a stream is observed, a :ref:`run <Stream>` message is sent “backwards” through the chain to the origin.
 
-As the message travels, it composes a :ref:`Sink` chain analogous to the stream chain. Finally, as the message reaches the origin, first then does the origin begin to produce its events. With the exception of a few combinators (such as :ref:`delay`), events propagate *synchronously* “forward” through the sink chain.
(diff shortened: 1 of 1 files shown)


docs/concepts.rst
```diff
@@ -59,14 +59,14 @@ Application error handling is outside the scope of these docs, as it is applicat
...
+Applying an operation to a stream derives a *new stream* with *new events*. There is *no alteration* to the original stream (*the origin*). Multiple operations compose a *chain of streams*. When a stream is observed, a :ref:`run <Stream>` message is sent “backwards” through the chain to the origin.
 
-As it travels, that message composes a :ref:`Sink` chain analogous to the source chain.  When the messages reaches the producer, it begins producing events.  With the exception of a few combinators (such as :ref:`delay`), events propagate *synchronously* "forward" through the sink chain.
+As the message travels, it composes a :ref:`Sink` chain analogous to the stream chain. Finally, as the message reaches the origin, first then does the origin begin to produce its events. With the exception of a few combinators (such as :ref:`delay`), events propagate *synchronously* “forward” through the sink chain.
```
davidchase: > first then does the origin begin to produce its events

is something missing from this pharse

Frikki: No, but we could make it better. E.g.:

> Finally, the origin begins to produce its events when the message reaches it.

davidchase: yeah that would be nice 👍

PR #165, opened 2017-10-26
Try lerna-changelog

briancavalier:
I figured I'd give [lerna-changelog](https://github.com/lerna/lerna-changelog) a try for comparison with #164.  I npm installed it, created a GitHub token, and then ran:

```
lerna-changelog --tag-from '@most/prelude@1.6.1' > CHANGELOG.md
```

To generate the changelog from an old tag without any additional configuration options.  It doesn't seem to be very well documented yet, although [babel uses it](https://github.com/babel/babel/blob/master/package.json#L44) and looking at the [lerna-changelog config in their lerna.json](https://github.com/babel/babel/blob/master/lerna.json#L4-L15) is helpful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,477 @@
+
+## @most/core@1.0.0-beta.0 (2017-10-25)
+
+#### :rocket: Enhancement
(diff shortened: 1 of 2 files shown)


briancavalier:
Closing for now, since this is out of date.  Still a good point of discussion, tho.

PR #167, opened 2017-10-29
Add hints about when and when not to use most/core

briancavalier:
We've seen folks using most/core in ways where it may not be a good fit, such as time series data processing.  I figured I'd start the discussion about how to make a short statement up front in the docs that might be helpful.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,14 +1,19 @@
-.. @most/core documentation master file, created by
-   sphinx-quickstart on Mon Jun 12 07:51:42 2017.
-   You can adapt this file completely to your liking, but it should at least
-   contain the root `toctree` directive.
-
 @most/core
 ==========
 
-``@most/core`` is a toolkit for reactive programming.  It helps you write highly interactive apps by composing streams of events, and without many of the hazards of side effects and mutable shared state.
+``@most/core`` is a toolkit for reactive event programming.  It helps you write highly interactive apps by composing :ref:`event streams <concepts>`, and without many of the hazards of side effects and mutable shared state.
 
 It features an ultra-high performance, low overhead architecture, and a small but powerful set of operations for merging, filtering, and transforming event streams.
+
+When should you use it?
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Use ``@most/core`` to build programs that transform, filter, coordinate, and *react to* :ref:`events` *as they happen*.  For example, it can be a good choice for building interactive user interfaces.
+
+When *shouldn't* you use it?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``@most/core`` is not a good choice for use cases that aren't characterized by reacting to events as they happen.  For example, it isn't intended for Array data processing, time series data analysis and processing, or for streaming chunked data with flow control (aka "back pressure").
 
 .. toctree::
    :maxdepth: 2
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -1,15 +1,20 @@
-.. @most/core documentation master file, created by
-   sphinx-quickstart on Mon Jun 12 07:51:42 2017.
-   You can adapt this file completely to your liking, but it should at least
-   contain the root `toctree` directive.
-
 @most/core
 ==========
 
-``@most/core`` is a toolkit for reactive programming.  It helps you write highly interactive apps by composing streams of events, and without many of the hazards of side effects and mutable shared state.
+``@most/core`` is a toolkit for reactive event programming.  It helps you write highly interactive apps by composing :ref:`event streams <concepts>`, and without many of the hazards of side effects and mutable shared state.
```
Frikki: Good one!

docs/index.rst
```diff
@@ -1,15 +1,20 @@
...
 
+When should you use it?
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Use ``@most/core`` to build programs that transform, filter, coordinate, and *react to* :ref:`events` *as they happen*.  For example, it can be a good choice for building interactive user interfaces.
+
+When *shouldn't* you use it?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``@most/core`` is not a good choice for use cases that aren't characterized by reacting to events as they happen.  For example, it isn't intended for Array data processing, time series data analysis and processing, or for streaming chunked data with flow control (aka "back pressure").
```
Frikki: **Wrong:** aka "“back pressure”

**Right:** a.k.a. “back pressure”

briancavalier:
Do you think trying to represent these two sections as bulleted lists would be more helpful than as paragraphs?

docs/index.rst
briancavalier: Wow, googling "aka acronym", "aka initialism", "aka vs a.k.a." yields a wide variety of info about which is "correct", and many seem to say with or without periods is equally acceptable.  There's even "a/k/a", which I'd never seen written until doing these searches!  Notably, the [Wikipedia Manual of Style](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Abbreviations) suggests _never_ using "aka", and [this article says the The Chicago Manual of Style](http://blogs.millersville.edu/bduncan/abbreviations-2/) indicates using "a.k.a."

Sigh ... I decided just to spell it out 😄

docs/index.rst
```diff
@@ -1,15 +1,20 @@
...
 
+When should you use it?
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Use ``@most/core`` to build programs that transform, filter, coordinate, and *react to* :ref:`events` *as they happen*.  For example, it can be a good choice for building interactive user interfaces.
+
+When *shouldn't* you use it?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``@most/core`` is not a good choice for use cases that aren't characterized by reacting to events as they happen.  For example, it isn't intended for Array data processing, time series data analysis and processing, or for streaming chunked data with flow control (also known as "back pressure").
```
Frikki: I believe it is just called “Array processing” and “time series analysis and processing”.

Frikki:
> Do you think trying to represent these two sections as bulleted lists would be more helpful than as paragraphs?

Not really. I find it hard to see how that would improve the separation between “should” and “shouldn’t”.

docs/index.rst
briancavalier: Updated in 7e561ab.

PR #177, opened 2017-11-12
Link and expand time types docs

### LINKED ISSUE #176: Improve documentation for time types
Improve documentation for time types, and link to it from functions that accept time types.

briancavalier:
Close #176

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -295,11 +295,9 @@ periodic
 
   periodic :: Period -> Stream void
 
-Create an infinite :ref:`Stream` containing events that occur at a specified :ref:`Period`. ::
+Create an infinite :ref:`Stream` containing events that occur at a specified :ref:`Period`.  The first event occurs at time 0, and the event values are ``undefined``. ::
 
   periodic(3): x--x--x--x-->
-
-The first event occurs at time 0, and the event values are ``undefined``.
 
 .. _throwError:
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -11,19 +11,48 @@ API
 Time
 ^^^^
 
+All time-related types use units defined by a :ref:`Clock`. The default :ref:`Scheduler` :ref:`Clock` uses milliseconds as its units: :ref:`Time`, :ref:`Delay <Delay-type>`, :ref:`Period`, and :ref:`Offset` will all be millisecond values.
+
 .. code-block:: haskell
 
   type Time = number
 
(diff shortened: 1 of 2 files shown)


docs/api.rst
```diff
@@ -11,19 +11,48 @@ API
 Time
 ^^^^
 
+All time-related types use units defined by a :ref:`Clock`. The default :ref:`Scheduler` :ref:`Clock` uses milliseconds as its units: :ref:`Time`, :ref:`Delay <Delay-type>`, :ref:`Period`, and :ref:`Offset` will all be millisecond values.
+
```
davidchase: if milliseconds are defined here do we need to repeat them for each one individually ?

briancavalier: There are probably better ways to structure things.  My thinking was that repeating it in multiple spots will be helpful.  I was considering two main ways people use documentation:

1. They read or skim parts of it like a book to understand concepts, or get a feel for the API, etc.
2. They look for the answer answer a specific question, like "what are the units of the number I pass to `delay`?"

Repetition is slightly annoying for number 1, but imho, very helpful for number 2.

For number 1, they'll likely encounter the `Time` heading first, and read that time units are milliseconds when using the default `Scheduler`.  My thinking is that they'll then just tend to ignore the repetition when they encounter it later.

For number 2, someone will tend just to go look for the `delay` function (possibly finding it by readthedocs search, or just find-on-page text search).  At that point they'll find a link to the `Delay` type, hopefully click on it, and see that its units are milliseconds.  If `Delay` doesn't say that it's milliseconds, the person would have to take another step: find the `Time` heading and read the intro prose.

docs/api.rst
```diff
@@ -266,7 +295,7 @@ periodic
 
   periodic :: Period -> Stream void
 
-Create an infinite :ref:`Stream` containing events that occur at a specified period. ::
+Create an infinite :ref:`Stream` containing events that occur at a specified :ref:`Period`. ::
```
Frikki: Should we add that the first event occurs at time 0?

briancavalier: In fact, it's specified twice just below: once in the ascii diagram, and once in the sentence below the ascii diagram.  Maybe the right thing is simply to move that lower sentence up to this first paragraph?  E.g.:

> Create an infinite :ref:`Stream` containing events that occur at a specified :ref:`Period`. The first event occurs at time 0, and the event values are undefined. ::

Frikki: Good idea.

briancavalier: Done in 3e90ef4

Issue #152, opened 2017-10-10
Return empty() when possible


briancavalier:
See #150 

There quite a few operations where if at least one input stream is empty(), the operation will always return a stream containing zero events. In such cases, some operations return the canonical empty(), and some don't.

It might be interesting to always return the canonical empty() in those cases. That would allow empty() to annihilate other operations more effectively. It trades some up-front cost for reduction in cpu and memory later when a stream is observed.

### Operations that could return empty()

* [x] `map`
* [x] `filter`
* [x] `ap`
* [x] `loop`
* [x] `withItems` & `zipItems` (previously implemented)
* [x] `zip` & `zipArray`
* [x] `combine` & `combineArray`
* [x] `mergeMapConcurrently`, and thus `join`, `chain`, and `concatMap`
* [x] `snapshot` & `sample` (#186)
* [x] `skipRepeats` & `skipRepeatsWith`
* [x] `slice`, `take`, `skip`
* [x] `takeWhile`
* [x] `skipWhile`
* [x] `skipAfter`
* ~`withLocalTime`~ - end time still needs to be localized, so `withLocalTime` can't be annihilated by `empty()`
* [x] `awaitPromises`
* [x] `switchLatest`
* [x] `throttle`
* [x] `debounce`
* [x] `merge(empty(), empty())` and `mergeArray` when all streams are `empty()`
* [x] `recoverWith(f, empty())`
* [x] `multicast`

### Operations that need more discussion

* `delay` - What is the _meaning_ of `delay(n, empty())`, especially wrt end?

### Related optimizations

Optimizations that don't produce canonical empty, but noting here so we don't lose them.

* `scan` - `scan(f, x, empty())` is `now(x)` rather than `empty()`. It's not clear whether this is a worthwhile optimization.

Frikki:
Do we have a list (or a point in the direction where to spot those places)?

briancavalier:
Nope.  We could start and maintain the list here in the original description, open a new PR for each item (or group of items), and check them off as we merge those PRs.  We'll just need to be a little careful about not letting this issue become one that stays open forever.

briancavalier:
I added a list in the description.

TylorS:
Couldn't `mergeArray` filter all streams out that are `empty()` ?

briancavalier:
> Couldn't mergeArray filter all streams out that are empty() ?

[...]

PR #179, opened 2017-11-20
Map return canonical empty

### LINKED ISSUE #178: Return canonical empty() in map operation

Reference #152

Frikki:
Close #178
Reference #152 

### DIFF
--- a/packages/core/src/fusion/Map.js
+++ b/packages/core/src/fusion/Map.js
@@ -6,6 +6,7 @@ import Pipe from '../sink/Pipe'
 import Filter from './Filter'
 import FilterMap from './FilterMap'
 import { compose } from '@most/prelude'
+import { isCanonicalEmpty } from '../source/empty'
 
 export default class Map {
   constructor (f, source) {
@@ -25,6 +26,10 @@ export default class Map {
    * @returns {Map|FilterMap} mapped source, possibly fused
    */
   static create (f, source) {
+    if (isCanonicalEmpty(source)) {
+      return source
+    }
+
     if (source instanceof Map) {
       return new Map(compose(f, source.f), source.source)
     }
--- a/packages/core/src/fusion/Map.js
+++ b/packages/core/src/fusion/Map.js
@@ -6,7 +6,7 @@ import Pipe from '../sink/Pipe'
 import Filter from './Filter'
 import FilterMap from './FilterMap'
 import { compose } from '@most/prelude'
-import { isCanonicalEmpty } from '../source/empty'
+import { isCanonicalEmpty, empty } from '../source/empty'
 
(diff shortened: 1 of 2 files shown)


packages/core/src/fusion/Map.js
```diff
@@ -25,6 +26,10 @@ export default class Map {
    * @returns {Map|FilterMap} mapped source, possibly fused
    */
   static create (f, source) {
+    if (isCanonicalEmpty(source)) {
+      return source
```
briancavalier: **tl;dr** I think returning `empty()` is the right thing here, and will be the right thing in many cases.

Thinking a bit more about this `isCanonicalEmpty` vs singleton business we've talked about on slack, I think letting types guide us can be helpful, even though we don't have types in the source yet.  If we think about the type of `map`:

```hs
map :: (a -> b) -> Stream a -> Stream b
```

The type of `source` is `Stream a`, and the return type of `map` is `Stream b`.  That means that, _in the general case_, `source`, which is of type `Stream a` can't be a valid return type of `map`, which returns `Stream b` (`source` can be a valid return value, for example, when `a` and `b` are the same concrete type, but that's a specific case, not the general one).

I think that argument is slightly loose right now, because the type of `empty` is currently documented as:

```hs
empty :: () -> Stream *
```

I'm not sure it's really clear what we meant by `*`.  For example, did we mean `any`, or did we mean "inferred"?  The [flow defs say `any`](https://github.com/mostjs/core/blob/master/packages/core/src/index.js.flow#L17).

So returning `empty()` will be ok: forall `b`, `any` is a valid `b`.  It probably will turn out to be the right thing in some others cases (`ap` comes to mind), but not all (it'd be possible to return `source` from `filter`, for example, because the types line up).

I realize there are probably spots in existing code that return the upstream source when they shouldn't.  Hopefully we can fix those as we go.

### Aside: strongly typed empty()

I don't think we need to do this now, but, we could look to Haskell's `List`, and it's empty variant for guidance:

```hs
data List a = Nil | Cons a (List a)
```

That means that `Nil` _still has the type_ `List a`.  Empty lists are still strongly typed, and you cannot pass an empty `List Int` to a function that accepts `List String`.

We may want to consider tightening the type of `empty` to:

```hs
empty :: () -> Stream a
```

A type system should infer `a` from what is being demanded by the caller (Flow does this.  I'm not sure about TS, but I suspect it does as well, given enough surrounding type hints).  I know the _implementation_ of `empty` currently uses a singleton, but like we talked about in slack, I feel like that really is an implementation detail.  For example, I can imagine an implementation of `empty` that returns a `new Empty()` on each call, and an implementation of `isCanonicalEmpty` that returns `instanceof Empty`.

PR #181, opened 2017-11-20
Filter return canonical empty

### LINKED ISSUE #180: Return canonical empty() in filter operation

Reference #152

Frikki:
Close #180
Reference #150, #152

### DIFF
--- a/packages/core/src/fusion/Filter.js
+++ b/packages/core/src/fusion/Filter.js
@@ -3,6 +3,7 @@
 /** @author John Hann */
 
 import Pipe from '../sink/Pipe'
+import { isCanonicalEmpty } from '../source/empty'
 
 export default class Filter {
   constructor (p, source) {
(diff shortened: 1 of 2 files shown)


packages/core/src/fusion/Filter.js
```diff
@@ -21,6 +22,10 @@ export default class Filter {
    * @returns {Filter} filtered source
    */
   static create (p, source) {
+    if (isCanonicalEmpty(source)) {
+      return source
```
briancavalier: Quick q: since we'll likely need to return `empty()` in the majority of other cases, should we return it here just for the sake of source code consistency?  (In terms of types and functionality, there's no difference)

Frikki: I was actually micro-optimizing here. Your call.

briancavalier: I'm cool with it as is.

PR #183, opened 2017-11-20
Applicative return canonical empty

### LINKED ISSUE #182: Return canonical empty() in ap operation

References #150, #152

Frikki:
Close #182
Reference #150, #152

### DIFF
--- a/packages/core/src/combinator/applicative.js
+++ b/packages/core/src/combinator/applicative.js
@@ -4,6 +4,7 @@
 
 import { combine } from './combine'
 import { apply } from '@most/prelude'
+import { isCanonicalEmpty } from '../source/empty'
 
 /**
  * Assume fs is a stream containing functions, and apply the latest function
@@ -16,5 +17,9 @@ import { apply } from '@most/prelude'
  * @returns {Stream} stream containing all the applications of fs to xs
  */
 export function ap (fs, xs) {
+  if (isCanonicalEmpty(xs)) {
+    return xs
+  }
+
   return combine(apply, fs, xs)
 }
--- a/packages/core/src/combinator/applicative.js
+++ b/packages/core/src/combinator/applicative.js
@@ -4,7 +4,7 @@
 
 import { combine } from './combine'
 import { apply } from '@most/prelude'
-import { isCanonicalEmpty } from '../source/empty'
+import { containsCanonicalEmpty, empty } from '../source/empty'
 
 /**
  * Assume fs is a stream containing functions, and apply the latest function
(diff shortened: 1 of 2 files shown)


packages/core/src/combinator/applicative.js
```diff
@@ -16,5 +17,9 @@ import { apply } from '@most/prelude'
  * @returns {Stream} stream containing all the applications of fs to xs
  */
 export function ap (fs, xs) {
+  if (isCanonicalEmpty(xs)) {
```
TylorS: Couldn't this optimization be extended to `fs` as well?

Frikki: I was thinking the same but wasn’t sure.

Frikki: @TylorS would you return `fs` if canonical empty?

TylorS: Yes, I would.

Frikki: So two if-statements returning each a canonical empty. Another approach is one if-statement returning neither of the original streams but `empty()`. I think that would fit @briancavalier’s statements about that he doesn’t care if the return canonical empty is the same as the given.

Frikki: So two if-statements returning each a canonical empty. Another approach is one if-statement returning neither of the original streams but `empty()`. I think that would fit @briancavalier’s statements about that he doesn’t care if the return canonical empty is the same as the given.

Frikki: So two if-statements returning each a canonical empty. Another approach is one if-statement returning neither of the original streams but `empty()`. I think that would fit @briancavalier’s statements about that he doesn’t care if the return canonical empty is the same as the given.

TylorS: :+1:

briancavalier: Because of types, I think specifically returning `empty()`, rather than returning either `xs` or `fs`, is the right thing (see [this comment](https://github.com/mostjs/core/pull/179#discussion_r152263187)).  So, I like the one `if`-statement idea: if `fs` or `xs` is canonical empty, return `empty()`.

packages/core/src/combinator/applicative.js
```diff
@@ -16,5 +17,9 @@ import { apply } from '@most/prelude'
  * @returns {Stream} stream containing all the applications of fs to xs
  */
 export function ap (fs, xs) {
+  if (containsCanonicalEmpty([fs, xs])) {
```
briancavalier: Clever 😄

Do you feel like this communicates intent more clearly than `if(isCanonicalEmpty(fs) || isCanonicalEmpty(xs)) {`?

briancavalier: Hmmm, now that I think about it, [`combine` was recently updated to detect canonical empty](https://github.com/mostjs/core/blob/master/packages/core/src/combinator/combine.js#L27).  We could potentially remove this check, but then _keep_ the new unit tests.

Thoughts?

Frikki: Oh, yes. We should definitely remove this check if the tests pass without it.

briancavalier: Sorry I missed that in earlier reviews!

PR #184, opened 2017-11-23
Correct the order of arguments in the sample combinator

JulesFrancoise:
The sample combinator was taking values from the sampler stream instead of the values stream


briancavalier:
Thanks @JulesFrancoise!

PR #190, opened 2017-11-27
Return canonical empty in zip, zipArray

### LINKED ISSUE #189: Return canonical empty in zip(), zipArray()
See #152

briancavalier:
Close #189

### DIFF
--- a/packages/core/src/source/empty.js
+++ b/packages/core/src/source/empty.js
@@ -8,6 +8,9 @@ export const empty = () => EMPTY
 export const isCanonicalEmpty = stream =>
   stream === EMPTY
 
+export const isNotCanonicalEmpty = stream =>
+  !isCanonicalEmpty(stream)
+
 class Empty {
   run (sink, scheduler) {
     return asap(propagateEndTask(sink), scheduler)
--- a/packages/core/src/source/empty.js
+++ b/packages/core/src/source/empty.js
@@ -8,8 +8,8 @@ export const empty = () => EMPTY
 export const isCanonicalEmpty = stream =>
   stream === EMPTY
 
-export const isNotCanonicalEmpty = stream =>
-  !isCanonicalEmpty(stream)
+export const containsCanonicalEmpty = streams =>
+  streams.some(isCanonicalEmpty)
 
 class Empty {
   run (sink, scheduler) {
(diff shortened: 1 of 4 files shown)


packages/core/src/source/empty.js
```diff
@@ -8,6 +8,9 @@ export const empty = () => EMPTY
 export const isCanonicalEmpty = stream =>
   stream === EMPTY
 
+export const containsCanonicalEmpty = streams =>
```
Frikki: I don’t think it is correct to introduce this functionality with this PR. If this commit would somehow later be reverted, or worse, dropped, subsequent functionality that _also_ relies_ on `containsCanonicalEmpty` would break.

briancavalier: My thinking was that this is exactly the functionality needed for the empty optimization in zip, and so I just decided to give it a name.  I could delete the function and call `streams.some` directly in zip.js instead.  That'd be ok with me.  In fact, I think I did it that way and then extracted the function.  However, I anticipate that exactly the same function can be used to implement a similar empty optimization for `combine` and `combineArray`.

Both flow and unit tests would notice the situation of removing `containsCanonicalEmpty` without also updating usages of it.

Thoughts on how to handle the situation?

Frikki: > However, I anticipate that exactly the same function can be used to implement a similar empty optimization for `combine` and `combineArray`.

That’s _exactly_ the reason why I don’t think it’s correct to introduce the function in this PR. Doing so would cause `combine` and `combineArray` to _depend_ on this PR never being reverted.

I’ve posted some valuable links regarding Git workflow in the Slack channel.

Frikki: OK. First of all, `containsCanonicalEmpty` is a _new_ feature that is introduced through the _empty_ module. It should at least be a separate commit, which would enable cherry-picking without “garbage”. Of course, if it deserves its own PR is subject to which Git workflow is followed. To me, the current workflow looks like a trunk-based development workflow, where _master_ is the trunk. Thus, a separate commit would suffice.

briancavalier: Gonna open a new PR to do this.  I'll commit the working zip/zipArray w/empty annihilation first, and containsCanonicalEmpty in a separate commit.

PR #204, opened 2017-12-08
Return canonical empty in takeWhile, skipWhile, and skipAfter

### LINKED ISSUE #203: Return canonical empty in takeWhile, skipWhile, skipAfter
See #152

briancavalier:
Close #203

### DIFF
--- a/packages/core/test/slice-test.js
+++ b/packages/core/test/slice-test.js
@@ -13,16 +13,16 @@ import { assertSame } from './helper/stream-helper'
 describe('slice', function () {
   describe('fusion', function () {
     it('should return empty given empty', () => {
-      is(empty(), slice(0, 1, empty()))
+      isCanonicalEmpty(slice(0, 1, empty()))
     })
 
     it('should return empty given zero-length slice', () => {
       const i = Math.floor(Math.random() * 1000)
-      is(empty(), slice(i, i, now(i)))
+      isCanonicalEmpty(slice(i, i, now(i)))
     })
 
     it('should return empty when fusion leads to zero-length slice', () => {
-      is(empty(), slice(1, 2, slice(1, 2, now(''))))
+      isCanonicalEmpty(slice(1, 2, slice(1, 2, now(''))))
     })
 
     it('should narrow when second slice is smaller', function () {
@@ -91,7 +91,7 @@ describe('slice', function () {
 
   describe('takeWhile', function () {
     it('given canonical empty stream, should return canonical empty', () => {
-      const s = takeWhile(_ => true, empty())
+      const s = takeWhile(Boolean, empty())
       assert(isCanonicalEmpty(s))
     })
 
@@ -109,7 +109,7 @@ describe('slice', function () {
 
   describe('skipWhile', function () {
     it('given canonical empty stream, should return canonical empty', () => {
-      const s = skipWhile(_ => false, empty())
+      const s = skipWhile(Boolean, empty())
       assert(isCanonicalEmpty(s))
     })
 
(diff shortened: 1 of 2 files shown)


packages/core/test/slice-test.js
```diff
@@ -90,6 +90,11 @@ describe('slice', function () {
   })
 
   describe('takeWhile', function () {
+    it('given canonical empty stream, should return canonical empty', () => {
+      const s = takeWhile(_ => true, empty())
```
Frikki: Since we don’t care whether `f` returns `true` or `false`, I think it would be clearer to use `Boolean`:

```ts
takeWhile(Boolean, empty())
```
Since we don’t care whether `f` returns `true` or `false`, I think it would be clearer to use `Boolean`:

```ts
takeWhile(Boolean, empty())
```

packages/core/test/slice-test.js
```diff
@@ -103,6 +108,11 @@ describe('slice', function () {
   })
 
   describe('skipWhile', function () {
+    it('given canonical empty stream, should return canonical empty', () => {
+      const s = skipWhile(_ => false, empty())
```
Frikki: Since we don’t care whether `f` returns `true` or `false`, I think it would be clearer to use `Boolean`:

```ts
takeWhile(Boolean, empty())
```

packages/core/test/slice-test.js
```diff
@@ -116,6 +126,11 @@ describe('slice', function () {
   })
 
   describe('skipAfter', function () {
+    it('given canonical empty stream, should return canonical empty', () => {
+      const s = skipAfter(_ => false, empty())
```
Frikki: Since we don’t care whether `f` returns `true` or `false`, I think it would be clearer to use `Boolean`:

```ts
takeWhile(Boolean, empty())
```

packages/core/test/slice-test.js
briancavalier: > we don’t care whether f returns true or false

Good point.  In fact, how about we make it completely arbitrary? I.e., return a random boolean to ensure we _can't_ care about it's value.

Frikki: I think what matters here is about making the test correct without unnecessarily halting the reader to try to figure out what’s going on, or why `f` returns `true` here and `false` in another place. As you have seen in my tests, I explicitly label my fixture setup. E.g.:

```ts
// Fixture setup
const arbitraryPredicate = _ => Math.random() >= 0.5 ? true : false
// Exercise system
const result = takeWhile(arbitraryPredicate, empty())
// Verify outcome
assert(isCanonicalEmpty(result))
```

Then the reader don’t have to examine the predicate or wonder about it. However, in this instance, I think that `Boolean` works just as well without explicit labels. On the other hand, if we make the random function ternary inline, it requires the reader to examine it.

briancavalier: I think `arbitraryPredicate` is pretty clear.  However, I wouldn't define it inline in the test--it could come from a test helper module.  I'm ok with `Boolean`, too, but I don't think it expresses the arbitrariness quite as explicitly.

Frikki: I meant inline as `takeWhile(_ => Math.random() >= 0.5 ? true : false, empty())`. Putting it in a helper module makes it less clear to the reader, should he decides to examine why a test is failing. For tests, I am a strong believer in code duplication for the sake of explicitness.

briancavalier: I still think `arbitraryPredicate` more clearly expresses the intent, and I don't see test helper modules as a problem.  That said, I also think `Boolean` is fine, and that it's more important to move this PR forward.

briancavalier: Switched to `Boolean` in 241c6fe.

PR #209, opened 2018-01-26
Port debounce fix from cujojs/most#515

briancavalier:
Post debounce fix from cujojs/most#515

See also https://github.com/cujojs/most/issues/514

### Todo

- [x] Unit test

### DIFF
--- a/packages/core/test/limit-test.js
+++ b/packages/core/test/limit-test.js
@@ -4,8 +4,12 @@ import { spy } from 'sinon'
 
 import { debounce, throttle } from '../src/combinator/limit'
 import { zip } from '../src/combinator/zip'
+import { startWith } from '../src/combinator/startWith'
+import { delay } from '../src/combinator/delay'
 import { map } from '../src/combinator/transform'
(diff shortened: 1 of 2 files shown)


packages/core/test/limit-test.js
```diff
@@ -65,6 +69,16 @@ describe('debounce', function () {
     })
   })
 
+  it('should not repeat last event when end is delayed', () => {
+    // See https://github.com/cujojs/most/issues/514
+    var expected = Math.random()
+    var end = delay(5, now(null))
+    var debounced = debounce(1, until(end, startWith(expected, never())))
+
+    return collectEventsFor(6, debounced)
```
Frikki: Why 6 ticks?

briancavalier: Lol, I think I was playing around with the test, and forgot to change it back to 5.

Also, `var` 😲

briancavalier: See ce1067b

Frikki: I didn't even notice var 😮

PR #216, opened 2018-03-19
POC: Add timeslice fusion

briancavalier:
Proof-of-concept implementation of fusion for timeslice operations (`until` and `since`).

This introduces an explicit time bounds type and associated operations for merging time bounds, and then reimplements `until` and `since` using them.  This allows adjacent timeslice operations to be fused by simply fusing their time bounds.

To avoid accumulating trees of disposables, this also adds "disposable fusion", which allows multiple disposables to be merged into a more efficient structure, or entirely replaced with DisposeNone.

### Todo

- [ ] Move "time bounds" type and operations to own module
- [ ] Add unit tests for time bounds operations
- [ ] Add unit tests for `until`/`slice` fusion
- [ ] Add unit tests for disposable merging
- [ ] Consider a similar "index bounds" type and operations to simplify fusing `skip`/`take`


packages/core/src/combinator/timeslice.js
```diff
@@ -3,111 +3,102 @@
...
-
-  run (sink, scheduler) {
-    const min = new Bound(-Infinity, sink)
-    const max = new UpperBound(this.maxSignal, sink, scheduler)
-    const disposable = this.source.run(new TimeWindowSink(min, max, sink), scheduler)
-
-    return disposeAll([min, max, disposable])
-  }
-}
+const timeslice = (bounds, stream) =>
```
briancavalier: This is the fusion algorithm.  It is essentially a right fold over adjacent `Timeslice`s, merging times bounds as it folds.

packages/core/src/combinator/timeslice.js
```diff
@@ -3,111 +3,102 @@
...
+const timeBounds = (min, minSignal, maxSignal) =>
+  ({ min, minSignal, maxSignal })
+
+const minBounds = (minSignal) =>
+  timeBounds(Infinity, minSignal, never())
+
+const maxBounds = (maxSignal) =>
+  timeBounds(0, never(), maxSignal)
+
+const mergeTimeBounds = (b1, b2) =>
```
briancavalier: Time bounds are a representation of the start/end bounds of a time slice:

* `min` is a known minimum time, before which events will be skipped.
* `minSignal` is a signal which can provide an update the minimum time
* `maxSignal` is a signal that represents the future maximum time after which events will be discarded, and the time sliced stream will end.

Merging time bounds takes the _later_ min value (i.e. min can only be "narrowed", and `Infinity` annihilates min values), and _merges_ signals.  Merging signals means the earliest event will win, which is what we want.

packages/disposable/src/disposeAll.js
```diff
@@ -1,9 +1,17 @@
...
 
 // Aggregate a list of disposables into a DisposeAll
-export const disposeAll = ds =>
-  new DisposeAll(ds)
+export const disposeAll = ds => {
+  const merged = ds.reduce(merge, [])
+  return merged.length === 0 ? disposeNone() : new DisposeAll(merged)
+}
+
+const merge = (ds, d) =>
```
briancavalier: `merge` merges a new disposable `d` into an array of disposables `ds`.  Since, disposeNone is a noop, it can be discarded instead of adding to the array of disposables.  If `d` is a DisposeAll (which contains an array of disposables), the two arrays, `ds` and `d.disposables` can just be concatenated.

packages/core/src/combinator/timeslice.js
```diff
@@ -3,111 +3,102 @@
...
+import { disposeBoth } from '@most/disposable'
 import { join } from './chain'
+import { merge } from './merge'
+import { never } from '../source/never'
+
+const timeBounds = (min, minSignal, maxSignal) =>
+  ({ min, minSignal, maxSignal })
+
+const minBounds = (minSignal) =>
+  timeBounds(Infinity, minSignal, never())
```
briancavalier: `Infinity` here is correct, but probably not obvious at first.  If you think about what a minimum bound _does_, then it's more obvious: we want to drop all events before we see the initial event from minSignal.  By starting with `Infinity`, we're saying that we want to discard all events before `Infinity`.  The first event in minSignal provides a new min (presumably earlier than `Infinity`!).

packages/core/src/combinator/timeslice.js
```diff
@@ -3,111 +3,102 @@
...
+
+const maxBounds = (maxSignal) =>
+  timeBounds(0, never(), maxSignal)
+
+const mergeTimeBounds = (b1, b2) =>
+  ({
+    min: Math.max(b1.min, b2.min),
+    minSignal: merge(b1.minSignal, b2.minSignal),
+    maxSignal: merge(b1.maxSignal, b2.maxSignal)
```
briancavalier: For these `merge`s to be maximally efficient, we could make `merge` discard `never` (as it does with `empty`).  I don't know that it matters all that much, since `never` is totally a noop stream (even moreso than `empty`!), and returns `disposeNone` which the disposable fusion (also in this PR) will handle efficiently.

briancavalier:
It probably makes sense to break this into 2 PRs:

1. Disposable fusion
2. Timeslice fusion (rebased on 1 after it's merged)

packages/core/src/combinator/timeslice.js
briancavalier: Hmmm, no, discarding `never` would not be correct for `merge`.  `never` would affect the end time of the resulting merged stream.  For example, while these contain the same event (value `1` at time zero), the first ends and the second doesn't: `now(1)` and `merge(now(1), never())`.

So, I think I'm content just to let the "noop-ness" of never, plus disposable fusion take care of the situation.

FWIW, Merge could squash multiple `never`s into a single `never`, but that doesn't seem worth it for the same reasons.  E.g. with disposable fusion, `mergeArray([never(), never(), never(), ...])` ends up just creating a bunch of `disposeNone`s, which disposable fusion would then squash into a single `disposeNone`.

packages/core/src/combinator/timeslice.js
briancavalier: > Merging signals means the earliest event will win, which is what I think we want.

Aha, wrong again!

That's correct for the maxSignal, but not for minSignal.  I think we want both only to narrow.  That is, I think we want the "earlier first event" of two maxSignals but the "later first event" of the two minSignals.

That's a bit trickier to implement.  Probably time for a little TDD to ensure I get that right.

briancavalier: Ok, 1809ca3 has an algorithm for narrowing minSignals (i.e. "latest first event").  The algorithm is lazy in that merging time bounds simply concatenates arrays of signal streams (a la free monoid).  Then, later in `Timeslice#run`, `getMinSignal` does the work of folding that array of min signals into a single stream containing the "latest first event" using `take` and `slice`.

packages/core/src/combinator/timeslice.js
```diff
@@ -4,110 +4,135 @@
...
+  timeBounds([min], [])
+
+const maxBounds = max =>
+  timeBounds([], [max])
+
+const mergeTimeBounds = (b1, b2) =>
+  timeBounds(b1.min.concat(b2.min), b1.max.concat(b2.max))
+
+// Interpret time bounds
+const getStartingMin = ({ min }) =>
```
Frikki: No reason to have a `get` prefix. `startingMin` is fine.

packages/core/src/combinator/timeslice.js
```diff
@@ -4,110 +4,135 @@
...
+  min.length === 0 ? 0 : Infinity
+
+const getSignals = ({ min, max }) =>
```
Frikki: Here we’re actually constructing a new object by applying operations to the input. An okay name would be `toSignals`.

packages/core/src/combinator/timeslice.js
briancavalier: :+1: I updated naming in e4edf53 based on your suggestions.

packages/core/src/combinator/timeslice.js
briancavalier: Agreed about dropping "get".  I went with `streamBounds` based on the observation that this function is: `FreeBounds (Stream a) -> Bounds (Stream a)`.

briancavalier:
@Frikki 

> (I think this PR adds extra bytes to the library.)

It certainly does.  It's adding a new feature: the ability to collapse adjacent timeslices into a single one.  So, the usual "does the feature warrant the bytes?" question applies.

davidchase:
general question should items like that be in this PR or is it better to open a separate new one? to keep this one small and easy to digest

briancavalier:
> items like that

Items like what?  Like the remaining todos?

davidchase:
yes items in todo

briancavalier:
Thanks for clarifying.  Yeah, I think all the remaining todos belong in this PR.  Without unit tests, imho, this PR wouldn't meet our (de facto) standards for being complete.  The other remaining todo is just to move existing code to its own module, which won't really change the scope of the PR.  Make sense?

davidchase:
yeah it does, i wasnt sure if adding testings and other files cluttered the PR even though are related but i gotcha

PR #211, opened 2018-02-13
docs: add initial ADR on external tooling

Frikki:
This PR is intended for receiving comments and suggestions. We could easily merge it, though not complete, as the document would just live as a proposal and can be updated to its development accordingly, leaving an excellent track record in Git.

Signed-off-by: Frederik Krautwald <fkrautwald@gmail.com>

PR #218, opened 2018-03-21
Add ADR template

briancavalier:
Add an ADR template, based on #211.


docs/adrs/TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
+# ADR 1: External Tooling
```
Frikki: Perhaps:

`# ADR [n]: [Title]`

...considering it is a template?

briancavalier: Good call, thanks.

PR #219, opened 2018-03-21
ADR: Add static types

briancavalier:
ADR for adding static types into the source.  Very interested to hear others' thoughts on pros and cons.

### DIFF
--- a/docs/adrs/001-add-types.md
+++ b/docs/adrs/001-add-types.md
@@ -1,0 +1,23 @@
+# ADR 1: Add Static Types
+
+Brian Cavalier (@briancavalier),
+David Chase (@davidchase),
+Tylor Steinberger (@TylorS)
+Frederik Krautwald (@Frikki),
+
+## Status
+
+**Proposed** | ~~Accepted~~ | ~~Deprecated~~ | ~~Superceded~~
+
+## Context
+
+We're already using Flow static types in some scheduler unit tests.  Adding static types to the source will increase the understandability, reduce mistakes, and enable encoding more structures explicitly (rather than keeping them in our heads or prose).
+
+Alternatives are:
+
+1. Flow. We're already using it in some unit tests.  It provides soundness, and has good tooling via VSCode and Atom.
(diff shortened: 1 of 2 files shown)


TylorS:
Does anyone know of places in the codebase where adding types would require a large refactoring?

briancavalier:
I'll bet there are some gunky place inside `@most/scheduler` where it might be tricky.  Otherwise, I don't think it will be too bad if we can go incrementally.  A lot of the code was written with types in mind, even though they aren't explicit.

I was thinking a good approach would be to start with `@most/disposable` and some key parts of `@most/core` like `slice`, `map`, and maybe `mergeConcurrently`.  With Flow, that should be relatively simple by incrementally adding `// @flow`.  Initially, that will only get us a little type safety, but as we add more coverage, obviously, we'll get more and more benefit.

Is it possible to do a similar incremental approach with TypeScript?  What are the mechanics of it? (incrementally renaming files to .ts?)

TylorS:
You can slowly convert to .ts files with the right tsconfig. With noImplicitAny and allowJs set to true I believe.

With just noImplicitAny true it's also pretty easy to mass rename to .ts since JS is valid TS.

briancavalier:
Thanks @TylorS.  I like the idea of renaming files incrementally so it's easy to tell at a glance which files have been converted and which haven't.  With Flow, we could tell by `// @flow` comments.

There's also [flow-coverage-report](https://www.npmjs.com/package/flow-coverage-report) (and just plain old `flow coverage`), as well as [type-coverage](https://www.npmjs.com/package/type-coverage) for TS.

nikoskalogridis:
My personal opinion would be to switch to using Flow types.

Advantages:

- Keeping the current build stack (babel, rollup) intact
- Completely static type check all the files in the code base
- Gradually migrate files (without renaming all of them in .ts)
- If we decide to change this decision and revert to plain Javascript we can simply strip the Flow types from the files and be done with it
- If we decide to switch to Typescript at some time in the future that will be easier as we would already have made most of the work required. We would only need to change some incompatibilities between Flow and Typescript.

briancavalier:
Ported to TypeScript in #296 🎉

PR #220, opened 2018-03-21
Improve slice bounds

briancavalier:
Extract slice bounds type and operations so they can be more easily verified.  Also, improve bounds algorithms and slice fusion in a few ways:

1. Rule out negatives to avoid potential min accumulation issues
2. Detect nil (min === max) and infinite bounds (0, Infinity)
3. Return canonical empty for nil bounds
4. Return input for infinite bounds

### DIFF
--- a/packages/core/src/combinator/slice/bounds.js
+++ b/packages/core/src/combinator/slice/bounds.js
@@ -1,5 +1,12 @@
 /** @license MIT License (c) copyright 2010 original author or authors */
 
+// A slice Bounds type that narrows min values via accumulation
+// and max values via Math.min.
+// 0 <= min <= max
+// slice(min2, max2 slice(min1, max1, s)) ~ slice(min1 + min2, Math.min(max1, max2), s)
+// type Bounds = { min: number, max: number }
+
+// smart constructor that enforces bounds constraints
 export const newBounds = (unsafeMin, unsafeMax) => {
   const min = Math.max(0, unsafeMin)
   const max = Math.max(min, unsafeMax)
@@ -16,8 +23,10 @@ export const maxBounds = max =>
 export const mergeBounds = (b1, b2) =>
   newBounds(b1.min + b2.min, Math.min(b1.max, b2.max))
 
+// nil bounds excludes all slice indices
 export const isNilBounds = b =>
   b.min >= b.max
 
+// infinite bounds excludes no slice indices
 export const isInfiniteBounds = b =>
   b.min <= 0 && b.max === Infinity
--- a/packages/core/src/combinator/slice/bounds.js
+++ b/packages/core/src/combinator/slice/bounds.js
@@ -3,25 +3,25 @@
 // A slice Bounds type that narrows min values via accumulation
 // and max values via Math.min.
 // 0 <= min <= max
-// slice(min2, max2 slice(min1, max1, s)) ~ slice(min1 + min2, Math.min(max1, max2), s)
+// slice(min2, max2, slice(min1, max1, s)) ~ slice(min1 + min2, Math.min(max1, max2), s)
 // type Bounds = { min: number, max: number }
 
 // smart constructor that enforces bounds constraints
-export const newBounds = (unsafeMin, unsafeMax) => {
+export const boundsFrom = (unsafeMin, unsafeMax) => {
   const min = Math.max(0, unsafeMin)
   const max = Math.max(min, unsafeMax)
   return { min, max }
 }
 
 export const minBounds = min =>
-  newBounds(min, Infinity)
+  boundsFrom(min, Infinity)
 
 export const maxBounds = max =>
-  newBounds(0, max)
+  boundsFrom(0, max)
 
 // Combine 2 bounds by narrowing min and max
 export const mergeBounds = (b1, b2) =>
-  newBounds(b1.min + b2.min, Math.min(b1.max, b2.max))
+  boundsFrom(b1.min + b2.min, Math.min(b1.max, b2.max))
 
 // nil bounds excludes all slice indices
 export const isNilBounds = b =>
--- a/packages/core/src/combinator/slice/bounds.js
+++ b/packages/core/src/combinator/slice/bounds.js
@@ -2,9 +2,15 @@
 
 // A slice Bounds type that narrows min values via accumulation
 // and max values via Math.min.
-// 0 <= min <= max
-// slice(min2, max2, slice(min1, max1, s)) ~ slice(min1 + min2, Math.min(max1, max2), s)
 // type Bounds = { min: number, max: number }
+// Notes:
+// 0 <= min <= max
+// slice(min2, max2, slice(min1, max1, s)) ~ slice(min1 + min2, Math.min(max1, min1 + max2), s)
+// A bounds has a 1d coord system with origin 0, extending to Infinity.  Both min and max
+// are relative to the origin (0).  However, when merging bounds b1 and b2, we
+// *interpret* b2 as being relative to b1, hence adding min1 to *both* min2 and max2.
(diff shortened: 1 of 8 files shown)


packages/core/test/combinator/slice/bounds-test.js
```diff
@@ -0,0 +1,75 @@
...
+  for (let i = 0; i < 100; ++i) {
+    const b = gen(i)
+    assert(b.min >= 0 && b.max >= b.min)
+  }
+}
+
+describe('slice/bounds', () => {
+  describe('newBounds', () => {
+    it('should create valid bounds', () => {
+      assertValidBounds(() => newBounds(genBound(), genBound()))
```
Frikki: There is a small possibility that this test could randomly fail and pass. If the subject under test’s source is modified actually to be a wrong implementation, the test could still pass should the 100 iterations result in a non-gaussian distribution.

packages/core/test/combinator/slice/index-test.js
```diff
@@ -27,14 +27,12 @@ describe('slice', function () {
 
     it('should narrow when second slice is smaller', function () {
       const s = slice(1, 5, slice(1, 10, now('')))
-      eq(2, s.min)
-      eq(6, s.max)
+      eq({ min: 2, max: 5 }, s.bounds)
```
Frikki: There is a changed behavior here because of how `mergeBounds`, used by `fuseSlice`, adds the minimums of the bounds, whereas the former implementation of `fuseSlice` calculated the fused bounds differently.

Could you please elaborate on this change?

packages/core/src/combinator/slice/bounds.js
```diff
@@ -0,0 +1,32 @@
+/** @license MIT License (c) copyright 2010 original author or authors */
+
+// A slice Bounds type that narrows min values via accumulation
+// and max values via Math.min.
+// 0 <= min <= max
+// slice(min2, max2 slice(min1, max1, s)) ~ slice(min1 + min2, Math.min(max1, max2), s)
+// type Bounds = { min: number, max: number }
+
+// smart constructor that enforces bounds constraints
```
Frikki: Using proper English in the comments is okay. 😄 Capitalize sentence and end with a period.

packages/core/src/combinator/slice/bounds.js
```diff
@@ -0,0 +1,32 @@
...
+export const newBounds = (unsafeMin, unsafeMax) => {
```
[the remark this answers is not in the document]
briancavalier: I like `boundsFrom` :+1:

packages/core/src/combinator/slice/bounds.js
```diff
@@ -0,0 +1,32 @@
...
+
+// Combine 2 bounds by narrowing min and max
+export const mergeBounds = (b1, b2) =>
+  newBounds(b1.min + b2.min, Math.min(b1.max, b2.max))
+
+// nil bounds excludes all slice indices
+export const isNilBounds = b =>
+  b.min >= b.max
+
+// infinite bounds excludes no slice indices
```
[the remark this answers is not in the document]
briancavalier: I was going for parallel language with `isNilBounds`'s comment, but it's probably less awkward to use your wording.  Thanks.

packages/core/src/combinator/slice/index.js
```diff
@@ -29,34 +28,35 @@ export const skip = (n, stream) =>
...
-  const fusedEnd = Math.min(end + sliceStream.min, sliceStream.max)
-  return slice(fusedStart, fusedEnd, sliceStream.source)
-}
+  sliceBounds(newBounds(start, end), stream)
+
+const sliceBounds = (bounds, stream) =>
+  isSliceEmpty(bounds, stream) ? empty()
+    : stream instanceof Map ? commuteMapSlice(bounds, stream)
+    : stream instanceof Slice ? fuseSlice(bounds, stream)
+    : createSlice(bounds, stream)
```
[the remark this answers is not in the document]
briancavalier: > They require an extra effort to understand

This seems subjective.  I find them to be very readable.

They're standard in the code base right now as well, so I'm inclined to keep this one.  I wouldn't be opposed to switching to `if`/`return`.  However, I feel it's something we should discuss first, given the current established ternary usage, and if we decide to do it, do it across the code base in another PR.  Sound ok?

packages/core/test/combinator/slice/bounds-test.js
```diff
@@ -0,0 +1,75 @@
...
+  maxBounds,
+  mergeBounds,
+  minBounds,
+  newBounds
+} from '../../../src/combinator/slice/bounds'
+
+const rInt = (min = 0, max = Number.MAX_SAFE_INTEGER) =>
+  min + (Math.random() * (max - min))
+
+const genBound = () =>
```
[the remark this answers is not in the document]
briancavalier: :+1:

packages/core/test/combinator/slice/bounds-test.js
briancavalier: Afaik, Math.random does generate a normal (gaussian) distribution.  So, this is effectively property testing over a normal distro.  We could up the iterations to 500 or 1000, or add a few exact tests specifically to test at the boundary conditions ... or both.

packages/core/src/combinator/slice/index.js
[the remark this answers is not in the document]
Frikki: Yup. I was stating my personal experience.

packages/core/test/combinator/slice/bounds-test.js
```diff
@@ -0,0 +1,75 @@
...
+import {
+  isInfiniteBounds,
+  isNilBounds,
+  maxBounds,
+  mergeBounds,
+  minBounds,
+  newBounds
+} from '../../../src/combinator/slice/bounds'
+
+const rInt = (min = 0, max = Number.MAX_SAFE_INTEGER) =>
```
Frikki: `rInt` doesn’t return an integer. A rounding function is needed.

packages/core/test/combinator/slice/bounds-test.js
```diff
@@ -0,0 +1,75 @@
...
+  min + (Math.random() * (max - min))
+
+const genBound = () =>
+  Math.random() >= 0 ? rInt() : -rInt()
```
Frikki: `Math.random()` is _always_ `>= 0`

packages/core/test/combinator/slice/bounds-test.js
Frikki: I have made a few tests, and it seems that 100 iterations produce a reasonably normal distribution most of the time.

packages/core/test/combinator/slice/index-test.js
briancavalier: Good catch.  I need to look into this, but haven't had time yet.

briancavalier:
@Frikki I addressed most of your feedback.  I still need to look into the 5 vs 6 change, tho.  Hopefully, I'll have time soon.

Frikki:
@briancavalier You’re doing an excellent job. I want to buy you a 🍺

packages/core/test/combinator/slice/index-test.js
briancavalier: Yep, this is a bug in the PR 👍.  I'm adding failing test cases ...

briancavalier: 5989444 adds some exact tests and fixes the bug.  I also added a comment to try to clarify how the algorithm works.

briancavalier:
Here's a [repl.it example you can play with](https://repl.it/repls/FlamboyantPrevailingFormats) that shows how the bounds merging algorithm is equivalent to a sequence of slice calls.

PR #223, opened 2018-04-12
add missing eslint dependencies

nikoskalogridis:


briancavalier:
Hi @nikoskalogridis, I apologize about the confusion here.  We have [switched to using standard](https://github.com/mostjs/core/blob/master/packages/prelude/package.json#L18), and `@most/eslint-config-most` is deprecated.  So, I think the right thing to do would be for us to delete the straggling `.eslintrc` files.  I'd be happy to do that, since it was a mess I made :) but if you'd rather do it in order to keep moving, that's fine too (and I'd be very thankful!).

nikoskalogridis:
@briancavalier sure I could remove it myself, no problem ;-)

briancavalier:
Thank you! I’d really appreciate it.

nikoskalogridis:
closing in favour of #226

PR #224, opened 2018-04-12
Allow merge to produce union

briancavalier:
As per recent discussion, this loosens the type of the 2-stream `merge` to accept streams of different types and produce a stream of the union.

At first, this might seem unsafe because you might accidentally call merge with different types, say `A` and `B` when you didn't mean to.  However, since the resulting stream has a union type, `A | B`, when you try to use it, for example by mapping it with a function `A => C`, the type checker will prevent it.  Thus, I believe this is entirely safe, while adding some flexibility for situations where you might actually want to construct a union.

### DIFF
--- a/packages/core/test/flow/combinator/merge.js
+++ b/packages/core/test/flow/combinator/merge.js
@@ -1,0 +1,10 @@
+// @flow
+import { merge, now, runEffects } from '../../../src/index'
+import { type Stream } from '@most/types'
+import { newDefaultScheduler } from '@most/scheduler'
+
+const s1 = now(123)
+const s2 = now('a')
+const s = merge(s1, s2)
+
+runEffects(s, newDefaultScheduler())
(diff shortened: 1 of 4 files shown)


packages/core/test/flow/combinator/merge.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+import { merge, now, runEffects } from '../../../src/index'
+import { type Stream } from '@most/types'
+import { newDefaultScheduler } from '@most/scheduler'
+
+const s1 = now(123)
+const s2 = now('a')
+const s = merge(s1, s2)
+
+runEffects(s, newDefaultScheduler())
```
Frikki: Strange test, but I guess it’s a type assertion test in the IDE. I’m not sure 🤔 it is okay, or do the node tests also check types?

Frikki: Ah, I see the path is `test/flow`

briancavalier: Yeah, see the readme in test/flow.  It'd probably be better to try to get typechecking working in merge's regular unit tests.  I tried briefly, but something must not be right in our flow config, since I couldn't get type checking to fail when it should :/

PR #225, opened 2018-04-12
Refactor VirtualTimer and expose it in scheduler package

nikoskalogridis:
Hello since this is my first commit please review carefully and let me know for any mess ups. As the title says I exposed `VirtualTimer` on the `@most/scheduler` package and updated the types (flow and typescript) and the docs.

Apart from the refactoring I did on `VirtualTimer` I have changed some behavior in it that I found puzzling at least for me. So please verify if the new behavior is the **wanted one**.

More specifically

- when we had a `setTimer()` with a `dt` > than the `dt` specified in the `tick()` invocation the timers `now()` time progressed at the `dt` of the `setTimer()` and the set function was also called.
- Also during the invocation of the `fn` given to the `setTimer` the `timer.now()` returned `Infinity`

both above behaviors are now changed so 

- calling `tick()` is going to result to forwarding the time at the `dt` specified in that invocation (calling the `fn` set with `setTimer` if its `dt` is passed)
- Also during invocation of a `setTimer` `fn` the `timer.now()` returns the correct time

As a test I replaced the old `VirtualTimer` on the `@most/core` tests and run the tests with success, so I guess that the behavior that I have changed was not something that a test was relying on.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -150,6 +150,19 @@ Timer
 
 A ``Timer`` abstracts platform time, typically relying on a :ref:`Clock`, and timer scheduling, typically using ``setTimeout``.
 
+.. _VirtualTimer:
+
+VirtualTimer
+^^^^^
+
+.. code-block:: haskell
+
+  type VirtualTimer = {
+    tick :: (Offset) -> void
+  }
+
+A ``VirtualTimer`` is a :ref:`Timer` with a tick function for progressing the Timer's time.
+
 .. _Timeline:
 
 Timeline
@@ -1371,6 +1384,24 @@ newClockTimer
   newClockTimer :: Clock -> Timer
 
 Create a new :ref:`Timer` that uses the provided :ref:`Clock` as a source of the current :ref:`Time`.
+
+.. _newVirtualTimer:
+
+newVirtualTimer
+`````````````
+
+.. code-block:: haskell
+
+  newVirtualTimer :: () -> VirtualTimer
+
+Create a new :ref:`VirtualTimer` that you can progress its time manually. Used mainly for testing.
+
+.. code-block:: javascript
+
+  const timer = newVirtualTimer()
+  console.log(timer.now()) // 0
+  timer.tick(10);
+  console.log(timer.now()) // 10
 
 .. _newTimeline:
 
--- a/packages/scheduler/src/VirtualTimer.js
+++ b/packages/scheduler/src/VirtualTimer.js
@@ -1,0 +1,60 @@
+/** @license MIT License (c) copyright 2010-2018 original author or authors */
+
+export default function newVirtualTimer () {
+  let nows = 0
+  let targetNow = 0
+  let task
+  let timer = null
+  let running = false
+  const key = {}
+
+  function stepTimer () {
+    if (nows >= targetNow) {
+      nows = targetNow
+      running = false
+      return
+    }
+    if (task !== undefined && targetNow >= task.time) {
+      nows = task.time
+    } else {
+      nows = targetNow
+    }
+    if (task !== undefined && nows === task.time) {
+      const fn = task.fn
+      task = undefined
+      if (typeof fn === 'function') {
+        fn()
+      }
+    }
+    timer = setTimeout(stepTimer, 0)
+  }
+
+  return {
+    now: () => nows,
+    setTimer: function (fn, dt) {
+      if (task !== undefined) {
+        throw new Error('VirtualTimer: Only supports one in-flight timer')
+      }
+      task = {fn, time: nows + Math.max(0, dt)}
+      return key
+    },
+    clearTimer: function (t) {
+      if (t !== key) {
+        return
+      }
+      clearTimeout(timer)
+      task = undefined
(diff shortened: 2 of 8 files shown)


briancavalier:
This looks really awesome, @nikoskalogridis.  I hope to make time to review it soon!

packages/scheduler/test/VirtualTimer-test.js
```diff
@@ -0,0 +1,62 @@
...
+        reject(new AssertionError('Timer triggered when it should not'))
+      }, dt + 1)
+      setTimeout(resolve(), 0)
+    })
+  })
+
+  it('should push current time to requested value if timer is set after that time', () => {
+    const timer = newVirtualTimer()
+    const dt = 243
+    const taskTime = dt + 10
```
Frikki: Could you factor out the `10` to, for example, `const arbitraryExcessTime = 10`? It is hard to understand what this magic number means.

packages/scheduler/test/VirtualTimer-test.js
```diff
@@ -0,0 +1,62 @@
...
+      setTimeout(resolve(), 0)
+    })
+  })
+
+  it('should not trigger timer when we do not tick manually', () => {
+    const timer = newVirtualTimer()
+    const dt = 234
+    return new Promise((resolve, reject) => {
+      timer.setTimer(() => {
+        console.log('triggered')
```
Frikki: Remove `console.log`.

packages/scheduler/test/VirtualTimer-test.js
```diff
@@ -0,0 +1,62 @@
...
+      timer.tick(dt)
+    })
+  })
+
+  it('should not trigger timer time upon ticking to an earlier time', () => {
+    const timer = newVirtualTimer()
+    const dt = 234
+    return new Promise((resolve, reject) => {
+      timer.setTimer(() => {
+        console.log('triggered')
```
Frikki: Remove console.log.

packages/scheduler/src/VirtualTimer.js
```diff
@@ -0,0 +1,60 @@
+/** @license MIT License (c) copyright 2010-2018 original author or authors */
+
+export default function newVirtualTimer () {
+  let nows = 0
```
Frikki: What is `nows`? When I see a name like that, I think of a list because of the plural. Wouldn’t `now` be fine?

docs/api.rst
```diff
@@ -1372,6 +1385,24 @@ newClockTimer
...
+
+  newVirtualTimer :: () -> VirtualTimer
+
+Create a new :ref:`VirtualTimer`. You progress its time manually, i.e., its time doesn’t progress automatically. Use it for example for testing.
+
+.. code-block:: javascript
+
+  const timer = newVirtualTimer()
+  console.log(timer.now()) // 0
+  timer.tick(10);
```
briancavalier: Very minor: remove semicolon.

Anyone know of a way to lint code samples in rST?

packages/scheduler/src/VirtualTimer.js
```diff
@@ -0,0 +1,60 @@
+/** @license MIT License (c) copyright 2010-2018 original author or authors */
```
briancavalier: We've started switching to a [new copyright date format](https://github.com/mostjs/core/blob/master/packages/core/src/combinator/slice/bounds.js#L1).  Not all files use it yet, so it's not obvious--sorry about that--but, as we touch files, we're trying to remember to update it.  The new format only mentions the starting copyright date:

```js
/** @license MIT License (c) copyright 2010 original author or authors */
```

packages/scheduler/test/VirtualTimer-test.js
```diff
@@ -0,0 +1,57 @@
...
+
+import newVirtualTimer from '../src/VirtualTimer'
+
+describe('VirtualTimer', () => {
+  it('new timer should start at time: 0', () => {
+    const timer = newVirtualTimer()
+    eq(0, timer.now())
+  })
+
+  it('should expose a tick function', () => {
```
briancavalier: I think this test is unnecessary: it is captured by the type VirtualTimer.

packages/types/index.js.flow
```diff
@@ -78,3 +78,7 @@ export type ScheduledTask = Disposable & {
   run (): void,
   error (Error): void,
 }
+
+export type VirtualTimer = Timer & {
+  tick (Offset): void
```
briancavalier: 👍 for using the Offset type here and in the TS.

docs/api.rst
```diff
@@ -150,6 +150,19 @@ Timer
 
 A ``Timer`` abstracts platform time, typically relying on a :ref:`Clock`, and timer scheduling, typically using ``setTimeout``.
 
+.. _VirtualTimer:
```
briancavalier: Thank you for including docs in the PR 😄

nikoskalogridis:
I think it would be a good idea to add two more functions on the VirtualTimer:

```
start(): void
stop(): void
```

Thoughts?

PR #227, opened 2018-04-13
remove deprecated inline eslint directives

nikoskalogridis:
sorry for missing them out on the previous PR

### DIFF
--- a/packages/core/src/fusion/Map.js
+++ b/packages/core/src/fusion/Map.js
@@ -14,7 +14,7 @@ export default class Map {
     this.source = source
   }
 
-  run (sink, scheduler) { // eslint-disable-line no-extend-native
+  run (sink, scheduler) {
     return this.source.run(new MapSink(this.f, sink), scheduler)
   }
 
--- a/packages/core/src/index.js
+++ b/packages/core/src/index.js
@@ -1,5 +1,4 @@
 /** @license MIT License (c) copyright 2016 original author or authors */
-/* eslint-disable import/first */
 import { curry2, curry3 } from '@most/prelude'
 
 export { empty } from './source/empty'
--- a/packages/core/src/invoke.js
+++ b/packages/core/src/invoke.js
@@ -3,7 +3,6 @@
 /** @author John Hann */
 
 export default function invoke (f, args) {
-  /* eslint complexity: [2,7] */
   switch (args.length) {
     case 0: return f()
     case 1: return f(args[0])
--- a/packages/core/src/LinkedList.js
+++ b/packages/core/src/LinkedList.js
@@ -27,7 +27,7 @@ export default class LinkedList {
    * Remove the provided node from the list
    * @param {{prev:Object|null, next:Object|null, dispose:function}} x node to remove
    */
-  remove (x) { // eslint-disable-line  complexity
+  remove (x) {
     --this.length
     if (x === this.head) {
       this.head = this.head.next
--- a/packages/core/src/combinator/slice/index.js
+++ b/packages/core/src/combinator/slice/index.js
@@ -73,7 +73,6 @@ class SliceSink extends Pipe {
   }
 
   event (t, x) {
-    /* eslint complexity: [1, 4] */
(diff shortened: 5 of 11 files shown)


nikoskalogridis:
interesting that the eslint directive was affecting standard .... anyway fixed it by reordering imports.
@briancavalier what is the reason of implementing curry inside the core/src/index.js and not exporting the functions already curried?

briancavalier:
Yeah, standard is eslint under the hood and does obey the comment directives.

The reason we only curry at the API boundary is so that internal code doesn't take the extra perf hit (small as it may be) of calling the curried versions.

packages/core/src/index.js
```diff
@@ -1,6 +1,34 @@
 /** @license MIT License (c) copyright 2016 original author or authors */
-/* eslint-disable import/first */
```
briancavalier: I'm in favor of not deleting this one, and keeping imports near their respective exports _in this particular file_.  I think it helps a good bit with readability.  Thoughts?

nikoskalogridis: Personally I am in favor of placing curry at the module boundary. 
That would allow the `index.js` to be readable and conform to standard with no overrides.
If we know that there is a significant performance benefit I guess that there is no option for that. 

In any case I don't like code standards overrides and bypasses (although I am not in favor of standard at all but since that decision is made I would like to avoid overrides and follow it all the time)

But if you feel strong about it I can revert this

nikoskalogridis: https://jsperf.com/curry-vs-nocurry/1

nikoskalogridis:
In order to not leave this PR handing while discussing I propose to open an issue for discussing the curry in the API or module boundary. So I reverted the reordering of imports in this PR. In any case if we decide to change the curry position `core/index.js` will change a lot

briancavalier:
Thanks for reverting the index.js reordering.  That'll help us merge this one more quickly!

packages/core/src/fusion/Map.js
```diff
@@ -14,7 +14,7 @@ export default class Map {
     this.source = source
   }
 
-  run (sink, scheduler) { // eslint-disable-line no-extend-native
+  run (sink, scheduler) {
```
briancavalier: 👍

packages/core/src/invoke.js
```diff
@@ -3,7 +3,6 @@
 /** @author John Hann */
 
 export default function invoke (f, args) {
-  /* eslint complexity: [2,7] */
```
briancavalier: Hmmm, I wonder if when we switched to standard we lost our `complexity` rule settings.  I think we previously had it intentionally set quite low, like 3 or 4--hence the explicit complexity comments.  If that's the case, then I think the right thing may be to re-enable our low complexity setting (if that's possible with standard), and keep or update all the complexity comments.

I'd like to hear from @TylorS, @Frikki, and @davidchase to see how they feel about the complexity rule as well.

Frikki: I feel the complexity rule is worthwhile when you have less experienced developers that could introduce unnecessary complexity. However, they would quickly resort to disabling the rule with a comment. Of course, this can be a good thing because attention will be brought to it during code reviews, but maybe we would spot the complexity anyway.

davidchase: iirc standard itself doesnt have any feelings towards complexity, however we can extend standard and add specific rules if we wanted. i think keeping complexity and at reasonable level would be good for us and any other contributor especially less experienced as @Frikki mentions 😄

briancavalier: I'm definitely in favor of re-introducing the complexity rule--seems like we lost it in the switch to standard.  We [had it set to 3](https://github.com/mostjs/eslint-config-most/blob/master/index.js#L11).  I like a very low setting as a default, so that we have to explicitly raise it or disable it in spots that are more complex.

What do you think?

Frikki: I am game.

davidchase: hooya

PR #228, opened 2018-04-14
add bind and noop functions

nikoskalogridis:
Since those are handy generic use functions and I use them in `@most/test` package that I am going to submit soon I think its the right place to put them

### DIFF
--- a/packages/prelude/src/function.js
+++ b/packages/prelude/src/function.js
@@ -47,3 +47,9 @@ export function curry4 (f) {
   }
   return curried
 }
+
+// noop :: () -> undefined
+export const noop = () => undefined
+
+// bind :: (a -> b) -> c -> (d -> e)
+export const bind = (fn, object) => fn.bind(object)
--- a/packages/prelude/test/function-test.js
+++ b/packages/prelude/test/function-test.js
@@ -3,7 +3,7 @@
 import {describe, it} from 'mocha'
 import assert from 'assert'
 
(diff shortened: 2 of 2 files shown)


packages/prelude/src/function.js
```diff
@@ -47,3 +47,9 @@ export function curry4 (f) {
   }
   return curried
 }
+
+// noop :: () -> undefined
+export const noop = () => undefined
+
+// bind :: (a -> b) -> c -> (d -> e)
+export const bind = (fn, object) => fn.bind(object)
```
briancavalier: Before we add this, I'd like to confirm that `bind`'s performance across VMs has improved.  Previously, it has been a significant performance hit, both when calling `fn.bind` and when calling previously-bound functions.

Frikki: Does performance here really matter? After all, it’s just a convenience API and not used internally.

briancavalier: I think it does matter.  The mostjs code never uses bind because of its history of performance issues--the current code just creates lambdas in spots it absolutely needs to.  If we add a `bind` function to prelude (and prelude is primarily for use in mostjs code), the intent is clearly that we will use it.  If that `bind` function also has performance issues, we could be introducing performance issues.

briancavalier:
If there are concrete use cases in mostjs libs for `noop`, I'm fine with adding it.  Are there?

I think to add `bind`, we'd need to see perf test results for the performance of Function.prototype.bind across current VMs.

PR #229, opened 2018-04-14
fix for reduce helper test function

nikoskalogridis:
while working on `@most/test` I found an error on this function. Thankfully it did not affect the validity of the tests. This solves two problems:

- Promise always returned `undefined` instead of the final value
- If the initial value was a primitive (ie. not passed by reference) the result would be the initial value

### DIFF
--- a/packages/core/test/helper/reduce.js
+++ b/packages/core/test/helper/reduce.js
@@ -2,7 +2,6 @@
 /** @author Brian Cavalier */
 /** @author John Hann */
 
-import Pipe from '../../src/sink/Pipe'
 import { runEffects } from '../../src/runEffects'
 import { tap } from '../../src/combinator/transform'
 import { newDefaultScheduler } from '@most/scheduler'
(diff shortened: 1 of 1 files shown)


briancavalier:
Oops!  Sorry for the breakage.  At one point we changed `end()` to always propagate undefined, since an end value had proven to be basically useless in practice.  Looks like we forgot to deal with all of the fallout.

packages/core/test/helper/reduce.js
```diff
@@ -13,10 +14,13 @@ import { newDefaultScheduler } from '@most/scheduler'
...
 * @param {Stream} stream to reduce
-* @returns {Promise} promise for the file result of the reduce
+* @returns {Promise} promise for the final result of the reduce
 */
-export const reduce = (f, initial, stream) =>
-  runEffects(new Reduce(f, initial, stream), newDefaultScheduler())
+export function reduce (f, initial, stream) {
+  let result = initial
+  const source = tap(x => { result = x }, new Reduce(f, initial, stream))
+  return runEffects(source, newDefaultScheduler()).then(() => result)
```
briancavalier: I wonder if we should just ditch Reduce and ReduceSink entirely, and just use tap (since this is only used for testing).  E.g. maybe something like:

```js
export function reduce (f, initial, stream) {
  let result = initial
  const source = tap(x => { result = f(result, x) }, stream)
  return runEffects(source, newDefaultScheduler()).then(() => result)
}
```

and then just delete Reduce and ReduceSink?

PR #232, opened 2018-04-16
Proposal: Sink event callbacks

nikoskalogridis:
I think it would be better to relax the return values of the sink callback functions and also expose them as types

Thoughts?

### DIFF
--- a/packages/types/index.d.ts
+++ b/packages/types/index.d.ts
@@ -12,9 +12,9 @@ export interface Stream<A> {
   run (sink: Sink<A>, scheduler: Scheduler): Disposable;
 }
 
-export type sinkEventCallback<A> = (time?: Time, value?: A) => mixed
-export type sinkEndCallback = (time?: Time) => mixed
-export type sinkErrorCallback = (time?: Time, error?: Error) => mixed
+export type sinkEventCallback<A> = (time?: Time, value?: A) => any
+export type sinkEndCallback = (time?: Time) => any
+export type sinkErrorCallback = (time?: Time, error?: Error) => any
 
 export interface Sink<A> {
   event: sinkEventCallback<A>;
--- a/packages/types/index.d.ts
+++ b/packages/types/index.d.ts
@@ -12,10 +12,14 @@ export interface Stream<A> {
   run (sink: Sink<A>, scheduler: Scheduler): Disposable;
 }
 
+export type sinkEventCallback<A> = (time?: Time, value?: A) => mixed
+export type sinkEndCallback = (time?: Time) => mixed
+export type sinkErrorCallback = (time?: Time, error?: Error) => mixed
+
 export interface Sink<A> {
-  event(time: Time, value: A): void;
-  end(time: Time): void;
-  error(time: Time, err: Error): void;
+  event: sinkEventCallback<A>;
+  end: sinkEndCallback;
+  error: sinkErrorCallback;
 }
 
 // Interface of a resource that can be disposed
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -79,9 +79,9 @@ Sink
 .. code-block:: haskell
 
   type Sink a = {
-    event :: Time -> a -> void
-    error :: Time -> Error -> void
-    end :: Time -> void
+    event :: Time -> a -> mixed
+    error :: Time -> Error -> mixed
+    end :: Time -> mixed
   }
 
 A ``Sink`` receives events—typically it does something with them, such as transforming or filtering them—and then propagates them to another ``Sink``.
--- a/packages/types/index.js.flow
+++ b/packages/types/index.js.flow
@@ -14,10 +14,14 @@ export type Stream<A> = {
   run (Sink<A>, Scheduler): Disposable
 }
 
+declare export function sinkEventCallback<A>(time?: Time, value?: A): mixed
+declare export function sinkEndCallback(time?: Time): mixed
+declare export function sinkErrorCallback(time?: Time, error?: Error): mixed
+
 export type Sink<A> = {
-  event (Time, A): void,
-  end (Time): void,
-  error (Time, Error): void,
+  event: sinkEventCallback<A>,
+  end: sinkEndCallback,
+  error: sinkErrorCallback,
 }
 
 // Interface of a resource that can be disposed


packages/types/index.d.ts
```diff
@@ -12,10 +12,14 @@ export interface Stream<A> {
   run (sink: Sink<A>, scheduler: Scheduler): Disposable;
 }
 
+export type sinkEventCallback<A> = (time?: Time, value?: A) => mixed
```
TylorS: TypeScript does not have a `mixed` type as Flow does. I believe `any` is the closest match that we would have without maybe defining our own "mixed" union type.

nikoskalogridis:
@TylorS, thanks I have already fixed those

TylorS:
This change looks good to me, thanks for taking the time! I'm curious to see what thoughts the other @mostjs/maintainers have

briancavalier:
@nikoskalogridis Could you explain what problem this solves, or what current/future use cases you imagine it supporting?

nikoskalogridis:
@briancavalier Since the sink callbacks are at the borderline of the API for this package I think that it would be a good idea to have them as type aliases and also loosen the requirements for their return values. That would allow a callback function taking no parameters or that returns a value to be used and also typecheck

davidchase:
@nikoskalogridis out of curiosity do you have a situation currently or perhaps one in mind where having those functions without parameters provide something that keeping them the way they are does not ?

davidchase:
@nikoskalogridis an example would be lovely, im trying to think of a situation where you would call any of those functions without either time and a value or an error.

nikoskalogridis:
For instance using flow:
```js
// @flow

type Time = number;

export type Sink<A> = {
  event (Time, A): void,
  end (Time): void,
  error (Time, Error): void,
}

const mySink: Sink<any> = {
  event: () => 4,
  error: () => 2,
  end: () => 4
}
```

This produces errors: 
```
12:   event: () => 4,
                   ^ Cannot assign object literal to `mySink` because number [1] is incompatible with undefined [2] in the return value of property `event`.
References:
12:   event: () => 4,
                   ^ [1]
6:   event (Time, A): void,
                      ^ [2]
13:   error: () => 2,
                   ^ Cannot assign object literal to `mySink` because number [1] is incompatible with undefined [2] in the return value of property `error`.
References:
13:   error: () => 2,
                   ^ [1]
8:   error (Time, Error): void,
                          ^ [2]
14:   end: () => 4                 ^ Cannot assign object literal to `mySink` because number [1] is incompatible with undefined [2] in the return value of property `end`.
References:
14:   end: () => 4                 ^ [1]
7:   end (Time): void,
                 ^ [2]
```

where with 
```
export type Sink<A> = {
  event (Time, A): mixed,
  end (Time): mixed,
  error (Time, Error): mixed
}
```

there are no errors


Apparently Typescript is more forgiving in the same scenario. So probably Typescript types can remain as is. 

Any Typescript expert can chime here with more knowledge about why this is so. But in any case I thought that keeping both flow and Typescipt aligned is a good idea so I have included the changes on the typings.d.ts file as well

davidchase:
thanks 😄  i think i got it with the typings im just trying to see an actually scenario where you would do such a thing?

nikoskalogridis:
You could pass a more general callback in there that in some cases return a value. In any case for me the question is rather why do we need this function to always return void. Do we depend on it inside the `@most` packages? Since it is not I would rather allow the users pass a function with a more loose type strictness.

Frikki:
@nikoskalogridis TypeScript’s `void` as a “return” type of a function, behaves differently than what we are generally used to in other type systems. In TypeScript, we can declare a function:

```typescript
interface Proc {
  (x: any): void
}
```

But our implementation can return _anything_ without producing a type error!

```typescript
const f: Proc = (x: any) => x + "Doh!"
```

However, if we call `f`:

```typescript
const x = f(3)
```

the type of `x` is `void` but its value is `"3Doh!".

See playground here:

https://www.typescriptlang.org/play/index.html#src=interface%20Proc%20%7B%0D%0A%20%20%20%20(x%3A%20any)%3A%20void%0D%0A%7D%0D%0A%0D%0A%2F%2F%20Unsound%20that%20we%20can%20return%20something.%0D%0Aconst%20f%3A%20Proc%20%3D%20(x%3A%20any)%20%3D%3E%20x%20%2B%20%22Doh!%22%0D%0A%0D%0A%2F%2F%20Sound%20that%20we%20must%20pass%20an%20argument.%0D%0A%2F%2F%20Sound%20that%20x%20is%20of%20type%20void.%0D%0Aconst%20x%20%3D%20f(3)%0D%0A%0D%0A%2F%2F%20Unsound%20that%20x%20of%20type%20void%20is%20actually%20a%20string.%0D%0A%2F%2F%20Logs%20%223Doh!%22%0D%0Aconsole.log(x)%0D%0A

briancavalier:
Thanks everyone for the discussion.  After thinking about this and reading the examples so far, I’m not in favor of relaxing these types.  I'll explain why.

1. Sink is one of the two primary core architectural component of mostjs.  It is used (almost) exclusively inside the mostjs architecture where I feel strongly that our policy should remain to try to be as rigorous as is possible/practical with types.  Type looseness compounds: being loose at a low level will inevitably propagate to higher levels (sometimes unknowingly since looseness can hide type errors).

2. The void return type makes a very strong and clear statement (at least in Flow): the type system prevents the architecture from doing anything (nefarious or otherwise) with return values because you literally _can't_ return anything except undefined.  We can't even fall back to dynamic JS and do runtime inspection on your return values.  The intent of mixed is far less clear: we might do something nefarious with your return values, or we might not.

2. I don’t perceive any problem that needs to be solved.  As far as I know, the `void` return hasn't caused the core team any pain when implementing mostjs features (not intending to speak for everyone--this is my current perception).

3. As far as we know, the community also does not perceive any sort of problem, i.e. the community also isn't feeling pain from the void return type, or isn't mentioning it.

For those reasons, I prefer keeping the more strict return type.

---

Now that we know about TS's `void` looseness, I'd even be in favor of finding a way to tighten the TS Sink return type.

Frikki:
@briancavalier 

Ad 1) I agree that loose types, and “poor” programming style in general, eventually will propagate to higher levels and infect the whole system.

Ad 2) Yes, the `mixed` return type indicates that we don’t know what we are dealing.

Ad 3) I also need to see a specific use case that absolutely **requires** the proposed change.

Ad 4) Well, the OP seem to perceive, maybe not a problem, but an issue. However, as mention in ad 3, I need to see a non-hypothetical use case whose solution cannot exist without this change.

---

I think that `void` in TypeScript _is_ the correct type for the `Sink` return. That the actual implementation can return something different doesn’t change how it can be used, i.e., whoever gets the returned value will be interpreted to be `void`. See the link in my comment above.

nikoskalogridis:
ok, I think that we have concluded regarding the return types of the callback functions. How about exporting the callback types named?

```js
declare export function sinkEventCallback<A>(time?: Time, value?: A): void
declare export function sinkEndCallback(time?: Time): void
declare export function sinkErrorCallback(time?: Time, error?: Error): void
```

do you think its a good idea? if so I can revert the changes about the return types and proceed with the named callbacks. Also is the naming convention correct? Are we ok in calling them as above or `sinkEventCb` etc would be better?

Also is there any point of declaring the `time, value, error` parameters as optional or that is not required?

nikoskalogridis:
@briancavalier no need to apologize. If someone should, that would be me. Thanks for your clarifications and for your detailed analysis of your thoughts. I think we can close this now :-)

briancavalier:
My pleasure, @nikoskalogridis.

@Frikki 

> I think that void in TypeScript is the correct type for the Sink return

Excellent point, thanks!  Let's leave it `void` in TS, since it's the right contract for callers.

PR #234, opened 2018-04-22
Proposal: Pipe function in prelude

nikoskalogridis:
thoughts?

### DIFF
--- a/packages/prelude/src/function.js
+++ b/packages/prelude/src/function.js
@@ -1,5 +1,7 @@
 /** @license MIT License (c) copyright 2010-2016 original author or authors */
 
+import { reduce } from './array'
+
 // id :: a -> a
 export const id = x => x
 
@@ -47,3 +49,6 @@ export function curry4 (f) {
   }
   return curried
 }
+
+const _pipe = (f, g) => (...args) => g(f(...args))
+export const pipe = (...fns) => reduce(_pipe, fns[0], fns.slice(1))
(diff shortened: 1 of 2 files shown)


davidchase:
I believe this maybe a duplicate of https://github.com/mostjs/core/pull/133

The reason `pipe` hasn't been added to prelude is because it isn't being used directly in the core since we have `compose`. 

With libraries like ramda, crocks, etc userland can easily create or import `pipe` and other functions. 

Thus we as contributors/maintainers do not have the burden of maintaining new functions and requests for adding new functions.

Out of curiosity what benefits do you see of adding `pipe`?

nikoskalogridis:
My thinking about this PR and also #228 is to include those functions since they are regularly used when writing in FP style and that both core and also other community created stream functions could use those without the need to bring in another library like ramda, reducing the number of dependencies.
I haven't seen #133 which indeed proposes the same function. In any case if you feel that it does not worth the trouble you can close this by all means :-)

briancavalier:
@nikoskalogridis One thing that may not be obvious--apologies--is that `@most/prelude` is intended for internal use in mostjs packages.  It's not really intended as a public, all-purpose lib of functional idioms.  We don't have the bandwidth to maintain something like that--plus there are several well-established libs already.  While it's appropriate for applications to depend on those libs, it didn't fit with mostjs's goals to depend on them, for reasons of library size and performance characteristics.

I'll close this issue since it's been discussed before.  As always, if mostjs internally ends up with a need for `pipe`, then we can consider adding it.

PR #235, opened 2018-04-23
Use denque-based Queue

### LINKED ISSUE #236: Update Queue to use techniques from Denque
See https://github.com/invertase/denque

briancavalier:
Update Queue to be based on [denque](https://github.com/invertase/denque) rather than [deque](https://github.com/petkaantonov/deque).  Apparently, denque is an updated approach that performs better on newer engines.

Here's a representative result:

```
❯ babel-node --presets env ./zip.js
zip 2 x 100000 integers
-------------------------------------------------------
most                118.98 op/s ±  0.87%   (71 samples)

~/Projects/cujojs/@most/core/packages/core/test/perf use-denque-zip* 3m 25s
❯ babel-node --presets env ./zip.js
zip 2 x 100000 integers
-------------------------------------------------------
most                135.29 op/s ±  0.53%   (72 samples)
```

I've run it numerous times, and the difference is consistently around 15%.  There's a small difference in library size, but unless you're actually using `zip` (or zipArray), Queue will be tree-shaken, so you won't notice.

I also updated the perf tests to get the all running again.

Edit (@Frikki):

Close #236


packages/core/test/perf/runners.js
```diff
@@ -65,6 +66,29 @@ function runSuite(suite) {
...
+
+const PropagateTask = require('../../src/scheduler/PropagateTask').PropagateTask
+const asap = require('../../../scheduler').asap
+
+function FromArray (a) {
+  this.array = a
+}
+
+FromArray.prototype.run = function (sink, scheduler) {
+  return asap(new PropagateTask(runProducer, this.array, sink), scheduler)
```
briancavalier: It's sufficient for the zip test, but we should write a custom FromArrayTask for the perf tests instead.  This current hacky approach will not work for anything that uses take or otherwise stops early.

briancavalier:
All the perf tests work again 😃

packages/core/src/Queue.js
```diff
@@ -1,64 +1,98 @@
...
+// Based on https://github.com/invertase/denque
 
 export default class Queue {
-  constructor (capPow2 = 32) {
-    this._capacity = capPow2
-    this._length = 0
+  constructor () {
     this._head = 0
+    this._tail = 0
+    this._capacityMask = 0x3
```
davidchase: what is this `0x3` what does that do?

briancavalier: Denque is implemented as a [ring buffer](https://en.wikipedia.org/wiki/Circular_buffer).  So, it needs to use mod arithmetic to find items in the buffer, since the actual head will not always be at position zero in `this._list`.  By carefully managing the size of the buffer, denque takes advantage of [this property of mod arithmetic](https://stackoverflow.com/questions/3072665/bitwise-and-in-place-of-modulus-operator), which allows using bitwise operations under those carefully controller conditions.  Bitwise `&` is more performant than mod `%` in JS (and on every other platform I know of).

Of course, the authors could have just written `3`, since that the same thing as `0x3`.  My guess is that they chose `0x3` and the name `_capacityMask` to emphasize that it is used in bitwise operations, like [this](https://github.com/invertase/denque/blob/master/index.js#L39).

FWIW, the current Queue implementation [exploits bitwise `&` for mod in exactly the same way](https://github.com/mostjs/core/blob/master/packages/core/src/Queue.js#L18).

davidchase: TIL 😃

packages/core/src/Queue.js
```diff
@@ -1,64 +1,98 @@
...
     const head = this._head
-    const x = this[head]
+    if (head === this._tail) {
+      return undefined
+    }
+
+    const x = this._list[head]
+    this._list[head] = undefined
+    this._head = (head + 1) & this._capacityMask
+    if (head < 2 && this._tail > 10000 && this._tail <= this._list.length >>> 2) {
```
Frikki: I think it would be useful to assign `10000` to a well-named variable or introduce a comment explaining why `10000`. Both would likely be worthwhile. Right now, `10000` is pure magic.

packages/core/src/Queue.js
```diff
@@ -1,64 +1,98 @@
...
-  _ensureCapacity (capacity) {
-    const oldCapacity = this._capacity
-    this._capacity = capacity
+  _shrinkArray () {
+    this._list.length >>>= 1
+    this._capacityMask >>>= 1
+  }
 
-    const last = this._head + this._length
+  _copyArray (fullCopy) {
```
Frikki: The underscore indicates a private method. `_copyArray` is only being called from _one_ place within the class with `fullCopy` set to `true`. I see no reason why L83 is making a check that _always_ will be _true_.

packages/core/test/perf/runners.js
```diff
@@ -65,6 +66,46 @@ function runSuite(suite) {
...
+
+class FromArrayTask {
+  constructor (array, sink) {
+    this.array = array
+    this.sink = sink
+    this.active = true
+  }
+
+  run (t) {
+    for (var i = 0, l = this.array.length; i < l && this.active; ++i) {
```
Frikki: If I get this right, `this.active` is `true` until `dispose` is called. Is it possible that `this.sink.event` can somehow trigger `dispose` of the instance to be called? If not, would it make sense to check `this.active` _before_ the for-loop to return early? Then `this.active` could also be removed on L97.

packages/core/src/Queue.js
briancavalier: Faster.  See the [representative result in the PR description](https://github.com/mostjs/core/pull/235#issue-183411386), and [denque's benchmarks](https://github.com/invertase/denque#benchmarks).

packages/core/src/Queue.js
briancavalier: I should probably explain my approach a bit more.  I copied the implementation of denque, and then retained only the methods we need (push and shift, and the ones they call) to keep code size down, and then adjusted the code style to match ours.  I wanted to change the actual implementation as little as possible for a couple reasons:

1. modifying the code of algorithms like this require a lot of care, since they rely on subtle things like bitwise ops, and
2. if there are changes to denque, we can incorporate them more easily

That said, I don't mind making very small tweaks, like extracting constants, if we feel it's worthwhile.

packages/core/test/perf/runners.js
briancavalier: > Is it possible that this.sink.event can somehow trigger dispose of the instance to be called?

It is: take() and takeWhile() will dispose a stream immediately when they detect their termination conditions.  That will happen from inside the call stack that starts at `this.sink.event`.  So, to be absolutely correct, and to allow take to be used in perf tests, the loop has to check `this.active` on every turn of the loop, and before calling `this.sink.end`.

Did that make sense?

packages/core/src/Queue.js
briancavalier: See https://github.com/mostjs/core/pull/235#discussion_r184567162.  I didn't want to change the denque code, if possible.  Happy to discuss other options, tho.

packages/core/src/Queue.js
Frikki: I feel `10000` is a magic number. My immediate questions are: Why `10000`? Why not `32768` or `16384`?

Thus, I think that an explanation of `10000` is appropriate in the code.

packages/core/src/Queue.js
Frikki: What are other options if you don’t want to change the code?

packages/core/test/perf/runners.js
Frikki: Yes, it does. All good.

packages/core/src/Queue.js
briancavalier: To be honest: my explanation is that I'm assuming the denque package has made the decision to use 10000 for a reason.  I've literally copied and pasted this code, only changing formatting :)

If we added denque as a dependency, we'd be in basically the same situation.  The reason I've not added it as a dependency is that we only need about 25% of the code in it, and it's prototype-based, so we'd have no choice but to end up with 100% of it in our build.

packages/core/src/Queue.js
briancavalier: Another option is to add denque as a dependency.  I feel like copying the code is a better solution in this case (see https://github.com/mostjs/core/pull/235#discussion_r185921969).

I do want to make sure we add proper attribution, though.  There's no license in their repo, though :/  We could open an issue to confirm they're ok with this approach as long as we include attribution.

davidchase: the package.json and badge both say MIT if that counts for anything

packages/core/src/Queue.js
Frikki: I think copy & modify is a good approach because we get to _own_ the code. Owning code is important because it gives absolute control. Now, since we at this point are in control, we should definitely document the `10000` magic number; at least as a constant.

packages/core/src/Queue.js
Frikki: In relation to this comment, https://github.com/mostjs/core/pull/235#discussion_r185966918, I also think we should optimize this function.

briancavalier:
@Frikki I extracted the two constants (2 and 10000) and wrote some comments about them.  I don't know why those particular values were chosen, so I just tried to document as much as I could based on how they were being used.  I cleaned up _copyArray as well.

packages/core/src/Queue.js
```diff
@@ -1,64 +1,106 @@
...
+// MIT License
 
-// Based on https://github.com/petkaantonov/deque
+// These constants were extracted directly from denque's shift()
+// It's not clear exactly why the authors chose these particular
+// values, but given denque's stated goals, it seems likely that
+// they were chosen for speed/memory reasons.
+
+// Max value of _head at which Queue is willing to shink
+// its internal array
```
Frikki: A complete sentence should end with a period. It is also applicable to code comments. The same applies below.

packages/core/src/Queue.js
```diff
@@ -1,64 +1,106 @@
...
+const HEAD_MAX_SHRINK = 2
+
+// Min value of _tail at which Queue is willing to shink
+// its internal array
+const TAIL_MIN_SHRINK = 10000
```
Frikki: We should make performance tests to establish this value, and also ask the origin why this value was chosen.

briancavalier: I feel like we can do these things without blocking this PR.  We're basically getting a 15% speed increase in zip for very little cost.

Frikki: Okay. Then we need to backlog that.

briancavalier: Added it to the [1.x Backlog project](https://github.com/mostjs/core/projects/4).

packages/core/src/Queue.js
Frikki: I still don’t see any reason to include the `fullCopy` parameter for the reasons mentioned in [my first comment](https://github.com/mostjs/core/pull/235/files/1685078eb2c627bc31649a37ab93516170cf25ef#r184538393).

packages/core/src/Queue.js
Frikki: Nah. I’m good as it is. :shipit:

PR #244, opened 2018-07-16
fix(delay) fixes the cancelAll predicate within DelaySink dispose

davidchase:
Init PR based on our slack conversation, addresses the bug described in #243


briancavalier:
It'd be great to have a failing test that this change fixes.  Creating a test based around the simplified example I posted in slack is probably good enough.  That said, I'd be fine merging this ASAP without the test, and following up in the next couple days with a test.

PR #248, opened 2018-07-17
Add npm release script

briancavalier:


briancavalier:
I'm guessing my changing "postinstall" to "bootstrap" broke the build.  I'll put that back and see what happens.

briancavalier:
Great suggestion, @Frikki.  Thanks!

PR #257, opened 2018-08-22
Discussion: add strawman most 2.0 package

briancavalier:
⚠️ DO NOT MERGE ⚠️ 

See: https://github.com/mostjs/core/projects/5

This is a strawman for discussing the possibility of providing a `most` 2.0 as a simple re-export of `@most/core` and friends.

Let's discuss the pros and cons to this kind of strategy.  It seems ok to include some discussion of the exact shape, but I'd rather not get _too_ wrapped up in that, and try to keep the discussion focused primarily on the overall strategy of "most 2.0 as a thin re-export / convenience layer".

### DIFF
--- a/packages/most/.flowconfig
+++ b/packages/most/.flowconfig
@@ -1,0 +1,17 @@
+[ignore]
+.*/node_modules/conventional-changelog-core/.*
+test/perf/.*
+
+[include]
+../../node_modules
+test/.*\.js
+test/flow/*.\.js
+dist/.*\.flow
+
+[libs]
+
+[options]
+include_warnings=true
+
+[lints]
+all=warn
--- a/packages/most/package-lock.json
+++ b/packages/most/package-lock.json
@@ -1,0 +1,3175 @@
+{
+  "name": "most",
+  "version": "2.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@most/core": {
+      "version": "1.3.2",
+      "resolved": "https://registry.npmjs.org/@most/core/-/core-1.3.2.tgz",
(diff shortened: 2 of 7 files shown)


briancavalier:
We had a great discussion on slack about a "no re-exports" alternative proposed by @TylorS based on his experience in [`@typed`](https://github.com/TylorS/typed).  Here's a summary of the idea and potential benefits:

### No re-exports

Create a `most` 2.0 package that lists `@most/core` et al. as dependencies, but _does not_ re-export anything.  The `most` 2.0 package would would have literally no exports.

(This relies on a flattened node_modules structure, as per npm >= 4 and yarn.  Node 8 is LTS and ships with npm 5, so this shouldn't be a practical concern.)

#### Pros compared to re-exports

* Trivial to shift between depending on individual packages or the `most` package, and thus, trivial upgrade path from `@most/core` to `most` 2.0 for devs who want that.
* Solves the problem of how to document _clearly_ the re-exports and individual packages: the packages you see in your code are always the packages that you see in the existing documentation.
* The path from “no re-exports” to “re-exports” is simple and non-breaking, if that turns out to be the better option.  The path from “re-exports” to “no re-exports” is breaking.

#### Cons compared to re-exports

* Can't import everything from a single package name.  This may feel inconvenient.
* Packages that appear in `import` statements are not the same as those listed in `package.json`.  This may be confusing.

briancavalier:
(More from slack)

#### Cons compared to re-exports

* Without re-exports, there may be an issue for users who do not have a build step, but , for example, wish to use the library directly in browser code.  They’d need to add `<script>` tags for all of the packages individually.
  * We felt this is something that we can solve if there's demand.

briancavalier:
Closing in favor of #258

Issue #266, opened 2019-01-31
unsubscribe equivalent


cezl:
Hi guys!
I was wondering if the @most/core has some sort of unsubscribe mechanism like most.js (cujojs) and RxJS. The closest operation to subscribe is run, but run (taking into consideration the similarities between Sink and observer), as specified in the docs, returns void. I'm asking that since I'm developing a project that allows to set (and afterwards to cancel/unsubscribe) a stream dynamically. I remember to have read in some place that @most/core avoids imperatives constructs and such, so what would be the alternative?

cezl:
Gotcha!
If that's the case, **I think** I could mix a [most-subject](https://github.com/mostjs-community/subject) subject with the combinator until in case I still want to manage the stream in a kind of imperative way :sweat_smile:

briancavalier:
Yes, absolutely.  One good general strategy is to represent the condition(s) under which disposal should occur as an event stream.  Then you can use `until`.  Of course, there are still use cases where `run`+`Disposable` may turn out to be the simplest thing :). Use your judgement, and feel free to discuss in [gitter](https://gitter.im/cujojs/most).

Recently, we've started recommending the newer [`@most/adapter`](https://github.com/mostjs/adapter) over `most-subject`.  If you go that route, be sure to check it out.

I feel like we've answered the original question of unsubscribe alternatives, so I'm going to close the issue. Please feel free to reopen if you feel there's a part of that original question that remains unanswered, and please do feel free to join us in gitter.  Thanks again for pointing out the documentation bug!

cezl:
hey @briancavalier, I have one more question.
First, I have accomplished what I wanted through the @most/adapter (I suggested most-subject first because I've only looked at the packages listed on [mostjs-community](https://github.com/mostjs-community)).
Second, while testing the following code with @most/adapter on Node.js, something called my attention:
```Javascript
const [induce, events] = createAdapter();

run({event: function(time, value){
    console.log(value);
},
error: function(time, error){
    console.error(error);
}, end: function(time){
    console.log('ended');
}}, newDefaultScheduler(), until(events, periodic(500)));

setInterval(()=> {
    induce(undefined);
}, 1550);
```
After the induce function run, the program didn't ended/stopped as I was expecting. For instance, the following code in RxJS would terminate the program as soon as the second interval emitted:
```Javascript
interval(500).pipe(takeUntil(interval(1550))).subscribe(console.log);
```
[...]

PR #270, opened 2019-01-31
Simplify until/since and fix subtle until dispose bug

briancavalier:
See [this comment](https://github.com/mostjs/core/issues/266#issuecomment-459485449) for an example of the bug.  It's a subtle interaction between `run` and `until`.

This PR fixes the bug.

Naively fixing the bug introduced a race between disposables.  SettableDisposable is the standard way to break disposable races in a controlled way, and it worked well here.

Looking at `since` and `until` with fresh eyes, it also seemed they could be simplified.  So, I did.  I also made Timeline slightly more aggressive in freeing its internal data structures when it knows it has an empty time slot.

### Todo

- [x] Add a unit test to prevent regression.  This is tricky to test, since it's in a part of the code that is hard to observe from the outside.

### DIFF
--- a/packages/core/src/combinator/timeslice.js
+++ b/packages/core/src/combinator/timeslice.js
@@ -3,8 +3,9 @@
 /** @author John Hann */
 
 import Pipe from '../sink/Pipe'
-import { disposeAll } from '@most/disposable'
+import { disposeBoth } from '@most/disposable'
 import { join } from './chain'
+import SettableDisposable from '../disposable/SettableDisposable'
 
 export const until = (signal, stream) =>
   new Until(signal, stream)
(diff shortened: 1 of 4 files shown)


briancavalier:
Added a test that times out on master, like the [original bug report](
https://github.com/mostjs/core/issues/266#issuecomment-459485449), and passes after the fix.  I think this is ready for a review.

packages/scheduler/src/ScheduledTask.js
```diff
@@ -19,6 +19,7 @@ export default class ScheduledTask {
   }
 
   dispose () {
+    this.active = false
```
briancavalier: This allows the scheduler to discard periodic tasks a bit earlier when they're canceled.

packages/scheduler/src/Timeline.js
```diff
@@ -26,6 +26,9 @@ export default class Timeline {
       const at = findIndex(st, this.tasks[i].events)
       if (at >= 0) {
         this.tasks[i].events.splice(at, 1)
+        if (this.tasks[i].events.length === 0) {
+          this.tasks.splice(i, 1)
+        }
```
briancavalier: Scheduler was holding onto empty future timeslots until their time arrived ... technically until the time slot after the empty one arrived.  So, this could have been leaking empty time slots in some rare cases.

packages/core/src/combinator/timeslice.js
```diff
@@ -37,77 +40,55 @@ class Since {
...
-    this.disposable = signal.run(this, scheduler)
+    this.disposable = disposable
   }
 
   event (t, x) {
-    if (t < this.value) {
-      this.value = t
-      this.sink.end(t)
-    }
+    this.disposable.dispose()
```
briancavalier: This explicit `dispose` is basically the crux of the fix, and is similar to what `take` does.

We may want to consider changing `run` to actually manage the final disposable in a similar way that runEffects does.  I think that would also have fixed this bug, but I felt like not fixing it here would leave `until` in a risky state.

PR #275, opened 2019-02-22
Add most-redis-docker example.

unscriptable:
I think it's hard to provide an example that is easy to run and is also somewhat close to a "real world" scenario.  By allowing users to run this example by installing docker _or_ installing redis, I hope to make it a bit easier.  

I'm very open to suggestions and improvements:  coding style, README contents, variable names, etc.  

Honestly, composing the most.js streams felt clunky without formal function composition tools.  Am I doing it right?  Could I do it better?

### DIFF
--- a/examples/most-redis-docker/.gitignore
+++ b/examples/most-redis-docker/.gitignore
@@ -1,2 +1,0 @@
-node_modules
-.DS_Store
--- a/examples/most-redis-docker/.gitignore
+++ b/examples/most-redis-docker/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+.DS_Store
--- a/examples/most-redis-docker/Dockerfile
+++ b/examples/most-redis-docker/Dockerfile
@@ -1,9 +1,6 @@
 FROM node:10.12.0-jessie
-# https://github.com/Yelp/dumb-init
-RUN wget -O /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64
-RUN chmod +x /usr/local/bin/dumb-init
-RUN mkdir -p /app
-WORKDIR /app
-ADD . /app/
+WORKDIR /usr/src/app
+COPY package*.json ./
 RUN npm install
-# CMD ["/usr/local/bin/dumb-init", "npm", "run", "start"]
+COPY . .
+# Entrypoint can vary so it must be provided at startup
--- a/examples/most-redis-docker/Dockerfile
+++ b/examples/most-redis-docker/Dockerfile
@@ -1,0 +1,9 @@
+FROM node:10.12.0-jessie
+# https://github.com/Yelp/dumb-init
+RUN wget -O /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64
+RUN chmod +x /usr/local/bin/dumb-init
+RUN mkdir -p /app
+WORKDIR /app
+ADD . /app/
+RUN npm install
+# CMD ["/usr/local/bin/dumb-init", "npm", "run", "start"]
--- a/examples/most-redis-docker/README.md
+++ b/examples/most-redis-docker/README.md
@@ -27,8 +27,9 @@ if you chose another folder name.
 
 This project uses environment variables.  If you are using the windows command
 prompt, you may have to change some commands slightly to use env vars. For
-example, rather than type `REDIS_MYCHANNEL=my_channel npm run consume`, you
-would type `set REDIS_MYCHANNEL=my_channel&&npm run consume`.
+example, rather than type
+`REDIS_MYCHANNEL=most-redis-docker-example npm run consume`, you would type
+`set REDIS_MYCHANNEL=most-redis-docker-example&&npm run consume`.
 
 ### Option 1: Docker
 
@@ -37,7 +38,7 @@ would type `set REDIS_MYCHANNEL=my_channel&&npm run consume`.
 2.  Build the docker image for this project:
 
 ```sh
-docker build -t most-redis .
+docker build -t most-redis-docker .
 ```
 
 ### Option 2: node.js
@@ -81,19 +82,19 @@ stop the docker-compose network in the background.
 1. From the `most-redis-docker` folder:
 
 ```sh
-REDIS_MYCHANNEL=my_channel npm run consume
+REDIS_MYCHANNEL=most-redis-docker-example npm run consume
 ```
 
 2. Open a new terminal or shell, and from the `most-redis-docker` folder type:
 
 ```sh
-REDIS_MYCHANNEL=my_channel npm run consume
+REDIS_MYCHANNEL=most-redis-docker-example npm run consume
 ```
 
 3. Open a third terminal or shell, and from the `most-redis-docker` folder type:
 
 ```sh
-REDIS_MYCHANNEL=my_channel npm run publish
+REDIS_MYCHANNEL=most-redis-docker-example npm run publish
 ```
 
 ## Experiment!
--- a/examples/most-redis-docker/README.md
+++ b/examples/most-redis-docker/README.md
@@ -112,7 +112,7 @@ Both the `consume` and `publish` services accept env vars.  If you are using
 Docker, these may be modified in the `docker-compose.yml` file.  The following
 env vars are used:
 
-- `REDIS_CHANNEL`: string, _required_
+- `REDIS_CHANNEL`: string, default is `'most-redis-docker-example'`
 - `REDIS_PORT`: integer, default is `6379`
 - `REDIS_HOST`: string, default is `'127.0.0.1'`
 - `TIMEOUT`: integer, default is `2500`
--- a/examples/most-redis-docker/README.md
+++ b/examples/most-redis-docker/README.md
@@ -28,8 +28,8 @@ if you chose another folder name.
 This project uses environment variables.  If you are using the windows command
 prompt, you may have to change some commands slightly to use env vars. For
 example, rather than type
-`REDIS_MYCHANNEL=most-redis-docker-example npm run consume`, you would type
-`set REDIS_MYCHANNEL=most-redis-docker-example&&npm run consume`.
+`REDIS_CHANNEL=most-redis-docker-example npm run consume`, you would type
+`set REDIS_CHANNEL=most-redis-docker-example&&npm run consume`.
 
 ### Option 1: Docker
 
@@ -89,19 +89,19 @@ redis-server
 2. From the `most-redis-docker` folder:
 
 ```sh
-REDIS_MYCHANNEL=most-redis-docker-example node consume
+REDIS_CHANNEL=most-redis-docker-example node consume
 ```
 
 3. Open a new terminal or shell, and from the `most-redis-docker` folder type:
 
 ```sh
-REDIS_MYCHANNEL=most-redis-docker-example node consume
+REDIS_CHANNEL=most-redis-docker-example node consume
 ```
 
 4. Open a third terminal or shell, and from the `most-redis-docker` folder type:
 
 ```sh
-REDIS_MYCHANNEL=most-redis-docker-example node publish
+REDIS_CHANNEL=most-redis-docker-example node publish
 ```
 
 ## Experiment!
--- a/examples/most-redis-docker/README.md
+++ b/examples/most-redis-docker/README.md
@@ -112,9 +112,9 @@ Both the `consume` and `publish` services accept env vars.  If you are using
 Docker, these may be modified in the `docker-compose.yml` file.  The following
 env vars are used:
 
+- `REDIS_CHANNEL`: string, _required_
 - `REDIS_PORT`: integer, default is `6379`
 - `REDIS_HOST`: string, default is `'127.0.0.1'`
-- `REDIS_CHANNEL`: string, required
 - `TIMEOUT`: integer, default is `2500`
 
 ### Try these scenarios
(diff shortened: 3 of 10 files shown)


examples/most-redis-docker/README.md
```diff
@@ -0,0 +1,124 @@
...
+would type `set REDIS_MYCHANNEL=my_channel&&npm run consume`.
+
+### Option 1: Docker
+
+1.  Install [docker](https://www.docker.com/get-started).  (You will need to
+    create a [DockerHub](https://hub.docker.com) account.)
+2.  Build the docker image for this project:
+
+```sh
+docker build -t most-redis .
```
briancavalier: What do you think of adding a `package.json` script for this? e.g. `npm run docker-build`.  Helpful?  Too much abstraction?

examples/most-redis-docker/README.md
```diff
@@ -0,0 +1,124 @@
...
+them.  In your terminal(s), you should expect to see informational logs from
+each service as they start up and as they process the JSON messages.
+
+### Option 1: Docker
+
+1.  From the `most-redis-docker` folder, create and run the docker-compose
+    network defined in `docker-compose.yml`:
+
+```sh
+docker-compose up
```
briancavalier: Similarly, would making `npm start` run this be helpful, or no?

examples/most-redis-docker/README.md
```diff
@@ -0,0 +1,124 @@
...
+Both the `consume` and `publish` services accept env vars.  If you are using
+Docker, these may be modified in the `docker-compose.yml` file.  The following
+env vars are used:
+
+- `REDIS_PORT`: integer, default is `6379`
+- `REDIS_HOST`: string, default is `'127.0.0.1'`
+- `REDIS_CHANNEL`: string, required
+- `TIMEOUT`: integer, default is `2500`
+
+### Try these scenarios
```
briancavalier: ❤️  I dunno if anyone will try these (hopefully!), but I love this section.

examples/most-redis-docker/config.js
```diff
@@ -0,0 +1,14 @@
...
+
+const config
+  = exports.config
+  = env => {
+    const port = Number(env.REDIS_PORT || 6379)
+    const host = String(env.REDIS_HOST || '127.0.0.1')
+    const timeout = Number(env.TIMEOUT || 2500)
+    const channel = env.REDIS_CHANNEL
+
+    if (!channel) throw new Error(`REDIS_CHANNEL not specified.`)
```
briancavalier: Since this is an example, how do you feel about making the channel default do something that's likely unique, like `most-redis-docker-example`, rather than failing?

examples/most-redis-docker/redis.js
```diff
@@ -0,0 +1,50 @@
...
+          client.on('ready', () => resolve(client))
+          client.on('error', reject)
+        }
+      )
+
+    // Create redis timeout detector (() => Promise)
+    const redisTimeout
+      = rejectAfter(timeout, `Redis failed to respond after ${timeout} msec!`)
+
+    return Promise.race([ clientP, redisTimeout() ])
```
briancavalier: Nice use of Promise.race 👍

briancavalier:
Sorry about the confusion wrt mostjs/examples repo vs mostjs/core examples dir.  I had kinda forgotten that both exist.  We need to figure out which one is canonical.

> composing the most.js streams felt clunky without formal function composition tools

Yeah, it's not as simple as one would like.  Variadic left-to-right composition, like ramda's pipe, helps.  There's also the experimental [`@most/fluent`](https://github.com/mostjs/x-fluent).  In the [type-to-search](https://github.com/mostjs/examples/blob/master/type-to-search/index.js#L22) example, we used the proposed ES pipeline operator, which is fairly pleasant, but requires a compile step.

examples/most-redis-docker/config.js
unscriptable: I'm on the fence.  On the one hand, it's very frustrating to get cryptic errors when trying to run an example.  

On the other hand, [early detection of configuration errors is important](https://blog.acolyer.org/2016/11/29/early-detection-of-configuration-errors-to-reduce-failure-damage/) for microservices.  I think it's also important for example apps:

- ✅ Defaulting to the wrong port or host will fail loudly and fast.  
- ✅ Defaulting to the wrong timeout will not affect 99.99% of the users.
- ❌ Defaulting to the wrong channel will fail silently, _causing user frustration_.

I'll create a `validate` function, inject a logging function, and emit a wordy, helpful message if the user skips the REDIS_CHANNEL.  I'll also provide a default so it doesn't crash.  I did not like how this `config` function wasn't separating concerns, anyways. 😄 

Sound good?

unscriptable: Ok, so the validate function got a bit unwieldy for an example app.  I just injected a logging function and used that instead of throwing.

briancavalier:
> mostjs/examples repo vs mostjs/core examples dir

Quick note specifically on this: the team discussed via slack and we're going with the mostjs/core examples dir as canonical.  So, this PR is in the right spot 🎉

examples/most-redis-docker/config.js
briancavalier: Good points.  I was thinking that for an example, it'd be ok to default.  I like the idea of adding logging in that case.  Ultimately, I'm cool with either:

* defaulting and logging which channel is being used (whether default or not), and highlighting when it has defaulted because of REDIS_CHANNEL not being specified.
* keeping the `throw` and updating the error message to say what the user needs to do to fix the issue.  Something like: "REDIS_CHANNEL not specified.  Please set the REDIS_CHANNEL env var to the name of a valid redis channel that this example should use, such as 'most-redis-docker-example'"

examples/most-redis-docker/README.md
briancavalier: Ah yes, good point.  I wasn't entirely tuned in to supporting both use cases.  This README is great, so I'm fine with letting docker users run docker commands directly.

examples/most-redis-docker/README.md
briancavalier: See other comment ... docker/docker-compose is ok.

> I wonder if we could start three instances of node from one `npm script`?

I'm sure there's a way ... although windows will probably cause problems.  Maybe something like [npm-run-all](https://www.npmjs.com/package/npm-run-all) can help.

examples/most-redis-docker/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Env vars
+
+Both the `consume` and `publish` services accept env vars.  If you are using
+Docker, these may be modified in the `docker-compose.yml` file.  The following
+env vars are used:
+
+- `REDIS_PORT`: integer, default is `6379`
+- `REDIS_HOST`: string, default is `'127.0.0.1'`
+- `REDIS_CHANNEL`: string, required
```
briancavalier: If we decide to keep REDIS_CHANNEL as required, what do you think of moving it first, and bolding or italicizing "_required_" to make it painfully obvious?

examples/most-redis-docker/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Experiment!
+
+### Env vars
+
+Both the `consume` and `publish` services accept env vars.  If you are using
+Docker, these may be modified in the `docker-compose.yml` file.  The following
+env vars are used:
+
+- `REDIS_CHANNEL`: string, _required_
```
unscriptable: REDIS_CHANNEL is now at the top to stand out better.

examples/most-redis-docker/config.js
```diff
@@ -0,0 +1,18 @@
...
+    const port = 'REDIS_PORT' in env ? Number(env.REDIS_PORT) : 6379
+    const host = 'REDIS_HOST' in env ? env.REDIS_HOST : '127.0.0.1'
+    const timeout = 'TIMEOUT' in env ? Number(env.TIMEOUT) : 2500
+    const channel
+      = 'REDIS_CHANNEL' in env ? env.REDIS_CHANNEL : 'most-redis-docker-example'
+
+    // The other env vars will cause obvious errors if missing, but
+    // `REDIS_CHANNEL` will cause silent failures if the user forgets to set it.
+    if (!('REDIS_CHANNEL' in env))
+      report(`Warning: REDIS_CHANNEL env var is missing. Using ${channel}.`)
```
unscriptable: Failure to specify REDIS_CHANNEL as an env var is more noticeable now.

examples/most-redis-docker/README.md
```diff
@@ -0,0 +1,132 @@
...
+`set REDIS_MYCHANNEL=most-redis-docker-example&&npm run consume`.
+
+### Option 1: Docker
+
+1.  Install [docker](https://www.docker.com/get-started).  (You will need to
+    create a [DockerHub](https://hub.docker.com) account.)
+2.  Build the docker image for this project:
+
+```sh
+docker build -t most-redis-docker .
```
unscriptable: I changed the name of the docker image (here and in docker-compose.yml) to better reflect the name of this example.

examples/most-redis-docker/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+.DS_Store
```
unscriptable: Is this file redundant now?  Should I remove it?

examples/most-redis-docker/README.md
```diff
@@ -0,0 +1,132 @@
...
+stop the docker-compose network in the background.
+
+### Option 2: node.js
+
+1.  Start redis.  See the [redis docs](https://redis.io/topics/quickstart) for
+    troubleshooting and options.
+
+```sh
+redis-server
+```
```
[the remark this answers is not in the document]
briancavalier: :+1:

examples/most-redis-docker/.gitignore
briancavalier: Good catch.  Yeah, I think it can be removed.

examples/most-redis-docker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM node:10.12.0-jessie
+WORKDIR /usr/src/app
+COPY package*.json ./
+RUN npm install
+COPY . .
+# Entrypoint can vary so it must be provided at startup
```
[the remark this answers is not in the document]
briancavalier: :+1:

examples/most-redis-docker/README.md
briancavalier: :+1:

examples/most-redis-docker/README.md
briancavalier: I guess technically speaking, REDIS_CHANNEL isn't required now.  Would you prefer to leave "_required_" here to encourage people to set it, or replace it with "default is ...."?

unscriptable: Fixing it...

examples/most-redis-docker/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+2. From the `most-redis-docker` folder:
+
+```sh
+REDIS_MYCHANNEL=most-redis-docker-example node consume
```
briancavalier: I just walked through the both the docker and non-docker instructions.  Docker worked perfectly.  Well done!

In the non-docker instructions, I found something I hadn't noticed before: The instructions use both names: `REDIS_MYCHANNEL`, and `REDIS_CHANNEL`.  The code wants `REDIS_CHANNEL`.  So, I think all we need to do is use update the docs to use `REDIS_CHANNEL` everywhere, and this whole thing is ready to 🚢.

screen shot 2019-02-28 at 6 46 47 am

screen shot 2019-02-28 at 6 52 44 am

UPDATE: The `env` in my screenshots is because I use [fish](https://fishshell.com) rather than a bash-compatible shell.  Keeping the the instructions optimized for bash-compatible is :+1:, as I imagine most fish users are accustomed to adapting.

unscriptable: Thanks.  I never would have noticed that!  Fixed.

examples/most-redis-docker/consume.js
```diff
@@ -0,0 +1,40 @@
...
+// Create a most.js stream
+const fromMessage
+  = map(message => JSON.parse(message).value)
+const logValue
+  = tap(value => log('received value =', value))
+
+const receiveNewValues
+  = fromMessage(logValue(messageStream))
+
+// TODO: transform or something?
```
Frikki: Could we put a better comment here? E.g., `// TODO: transform the stream as a useful exercise.`

examples/most-redis-docker/consume.js
```diff
@@ -0,0 +1,40 @@
...
+runEffects(receiveNewValues, newDefaultScheduler())
+
+client
```
Frikki: A comment showing intent would be helpful here.

examples/most-redis-docker/publish.js
```diff
@@ -0,0 +1,39 @@
...
+  = scan(b => b + 1)
+const toJsonMessage
+  = map(value => JSON.stringify({ value }))
+const logValue
+  = tap(value => log('sending value =', value))
+
+const emitPeriodicMonotonicNumbers
+  = logValue(toJsonMessage(monotonicallyIncrease(0)(everyThreeSeconds)))
+
+client
```
Frikki: Again, put a useful CSI here, so the reader knows why we run this code.

examples/most-redis-docker/consume.js
unscriptable: Good catch, @Frikki.  I had looked past them so many times, they were invisible to me. Check out the comments I added.

briancavalier:
🎉 Thanks, everyone!

TylorS:
Thank you for this example @unscriptable !

unscriptable:
Got any other ideas for server-side examples? 😄

TylorS:
I feel anything with web sockets would make a good example - if that sounds interesting to you 😄 

Maybe we should make some more issues to track the examples move into this repo and discuss some possible example additions we could create.

PR #281, opened 2019-05-06
Apply new technique for strong heterogenous array types

briancavalier:
This uses a variation of the technique from [Simpler Variadic Function Types in Typescript](https://gist.github.com/briancavalier/62f784e20e4fffc4d671126b7f91bad0) to improve the types of zipArray and combineArray for TS, allowing them to scale to any arity.

It also improves the type of mergeArray, allowing it to be heterogeneous while remaining strongly typed.  Essentially, this extends the union behavior of `merge` to arbitrary arrays of streams.  This works by leveraging the fact that TS union types obey a distributive law.

### Potential Gotchas

In general, I see strong heterogeneous types as a plus.  However, there could be a couple downsides:

1. They could be surprising or confusing for some users.
2. I haven't found any way to achieve parity in Flow.  What are the implications for documentation?

### Todos

- [x] `ToStreamsArray` is duplicated, consider extracting it somewhere
- [x] Figure out the implications for docs, given that Flow can't handle this (afaict)

### DIFF
--- a/packages/core/src/index.js.flow
+++ b/packages/core/src/index.js.flow
@@ -76,7 +76,7 @@ declare export function switchLatest <A> (s: Stream<Stream<A>>): Stream<A>
 declare export function merge <A, B> (s1: Stream<A>, s2: Stream<B>): Stream<A | B>
 declare export function merge <A, B> (s1: Stream<A>): (s2: Stream<B>) => Stream<A | B>
 
-type MergeArray = <A>(Array<Stream<A>>) => Stream<A>
+type MergeArray = <A>($ReadOnlyArray<Stream<A>>) => Stream<A>
 declare export function mergeArray <S> (ss: S): $Call<MergeArray, S>
 
 declare export function sample <A> (values: Stream<A>, sampler: Stream<any>): Stream<A>
--- a/packages/core/src/index.js.flow
+++ b/packages/core/src/index.js.flow
@@ -75,7 +75,9 @@ declare export function switchLatest <A> (s: Stream<Stream<A>>): Stream<A>
 
 declare export function merge <A, B> (s1: Stream<A>, s2: Stream<B>): Stream<A | B>
 declare export function merge <A, B> (s1: Stream<A>): (s2: Stream<B>) => Stream<A | B>
-declare export function mergeArray <A> (ss: Array<Stream<A>>): Stream<A>
+
+type MergeArray = <A>(Array<Stream<A>>) => Stream<A>
+declare export function mergeArray <S> (ss: S): $Call<MergeArray, S>
 
 declare export function sample <A> (values: Stream<A>, sampler: Stream<any>): Stream<A>
 declare export function sample <A> (values: Stream<A>): (sampler: Stream<any>) => Stream<A>
--- a/packages/core/type-definitions/combinator/combine.d.ts
+++ b/packages/core/type-definitions/combinator/combine.d.ts
@@ -1,14 +1,10 @@
 import { Stream } from '@most/types';
+import { ToStreamsArray } from './variadic'
 
 export function combine<A, B, R>(fn: (a: A, b: B) => R, a: Stream<A>, b: Stream<B>): Stream<R>;
 export function combine<A, B, R>(fn: (a: A, b: B) => R): (a: Stream<A>, b: Stream<B>) => Stream<R>;
 export function combine<A, B, R>(fn: (a: A, b: B) => R, a: Stream<A>): (b: Stream<B>) => Stream<R>;
 export function combine<A, B, R>(fn: (a: A, b: B) => R): (a: Stream<A>) => (b: Stream<B>) => Stream<R>;
-
-// TODO: use readonly any[] once TS 3.4.x has been in the wild for "enough" time
-type ToStreamsArray<A extends ReadonlyArray<any>> = {
-  [K in keyof A]: Stream<A[K]>
-}
 
 export function combineArray<Args extends any[], R>(fn: (...args: Args) => R, streams: ToStreamsArray<Args>): Stream<R>;
 export function combineArray<Args extends any[], R>(fn: (...args: Args) => R): (streams: ToStreamsArray<Args>) => Stream<R>;
--- a/packages/core/type-definitions/combinator/combine.d.ts
+++ b/packages/core/type-definitions/combinator/combine.d.ts
@@ -5,6 +5,7 @@ export function combine<A, B, R>(fn: (a: A, b: B) => R): (a: Stream<A>, b: Strea
 export function combine<A, B, R>(fn: (a: A, b: B) => R, a: Stream<A>): (b: Stream<B>) => Stream<R>;
 export function combine<A, B, R>(fn: (a: A, b: B) => R): (a: Stream<A>) => (b: Stream<B>) => Stream<R>;
 
+// TODO: use readonly any[] once TS 3.4.x has been in the wild for "enough" time
 type ToStreamsArray<A extends ReadonlyArray<any>> = {
   [K in keyof A]: Stream<A[K]>
 }
--- a/packages/core/type-definitions/combinator/combine.d.ts
+++ b/packages/core/type-definitions/combinator/combine.d.ts
@@ -5,39 +5,9 @@ export function combine<A, B, R>(fn: (a: A, b: B) => R): (a: Stream<A>, b: Strea
 export function combine<A, B, R>(fn: (a: A, b: B) => R, a: Stream<A>): (b: Stream<B>) => Stream<R>;
 export function combine<A, B, R>(fn: (a: A, b: B) => R): (a: Stream<A>) => (b: Stream<B>) => Stream<R>;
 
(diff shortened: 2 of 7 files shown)


packages/core/type-definitions/combinator/combine.d.ts
```diff
@@ -5,39 +5,9 @@ export function combine<A, B, R>(fn: (a: A, b: B) => R): (a: Stream<A>, b: Strea
...
-): Stream<R>;
-export function combineArray<A, B, C, D, E, R>(
-  fn: (a: A, b: B, c: C, d: D, e: E) => R,
-  streams: [Stream<A>, Stream<B>, Stream<C>, Stream<D>, Stream<E>]
-): Stream<R>;
-export function combineArray<V, R> (
-  fn: (...items: V[]) => R,
-  items: Stream<V>[]
-): Stream<R>;
+type ToStreamsArray<A extends ReadonlyArray<any>> = {
```
Frikki: Consider `readonly any[]`.

packages/core/type-definitions/combinator/merge.d.ts
```diff
@@ -2,4 +2,8 @@ import { Stream } from '@most/types';
 
 export function merge<A, B>(s1: Stream<A>, s2: Stream<B>): Stream<A | B>;
 export function merge<A, B>(s1: Stream<A>): (s2: Stream<B>) => Stream<A | B>
-export function mergeArray<A>(streams: Array<Stream<A>>): Stream<A>;
+
+type MergeArray<S extends ReadonlyArray<Stream<any>>> = Value<S[number]>
```
Frikki: Consider `readonly Stream<any>[]`

packages/core/type-definitions/combinator/merge.d.ts
```diff
@@ -2,4 +2,8 @@ import { Stream } from '@most/types';
...
+type Value<S> = S extends Stream<infer A> ? A : never
+
+export function mergeArray<S extends ReadonlyArray<Stream<any>>>(streams: S): Stream<MergeArray<S>>;
```
Frikki: Consider `readonly Stream<any>[]`

packages/core/type-definitions/combinator/zip.d.ts
```diff
@@ -5,38 +5,9 @@ export function zip<A, B, R>(fn: (a: A, b: B) => R): (a: Stream<A>, b: Stream<B>
...
-): Stream<R>;
-export function zipArray<A, B, C, D, E, R>(
-  fn: (a: A, b: B, c: C, d: D, e: E) => R,
-  streams: [Stream<A>, Stream<B>, Stream<C>, Stream<D>, Stream<E>]
-): Stream<R>;
-export function zipArray<V, R> (
-  fn: (...items: V[]) => R,
-  items: Stream<V>[]
-): Stream<R>;
+type ToStreams<A extends ReadonlyArray<any>> = {
```
Frikki: Consider `readonly any[]`

packages/core/type-definitions/combinator/merge.d.ts
briancavalier: Ah, I do like that.  My only concern is compatibility.  Do you know in which TS version that syntax was introduced?  If it's relatively new, we may want to wait a minor version or two, but if it's already been around a while, we can probably go with it.

briancavalier: Looks like it was introduced [in 3.4.0](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html).  Thoughts on waiting until 3.5.0 is released vs using `readonly` now while 3.4.x is the latest minor?

Frikki: Since it was released with v3.4.0, a non-breaking feature release, I don’t see any reason to delay the use. I mean, what would be the point of waiting for v3.5.0? If we delay the application, then, as a minimum, I think we should add TODO comments.

briancavalier: I'm thinking more in terms of breaking compatibility with anyone using TS < 3.4.  We'd be introducing a type that breaks their build.  This seems like a slightly gray area for semver, I guess strictly speaking, we'd need to release a new major version of `@most/core`.

I like the TODO idea, or even create a github issue now to switch from `ReadonlyArray` to `readonly`.

unscriptable:
I’m going to try to find some time to investigate this technique in Flow.  I’ve been looking for a generic solution for this pattern for a while.

briancavalier:
I'd love to find at least an approximation in Flow.  I tried (very) briefly to get something working with type-level functions (via $Call), but never managed to find something I was really happy with.  It'll be great to have someone else give it a go!

briancavalier:
@unscriptable Here's the gist of what I was [fiddling with](https://flow.org/try/#0C4TwDgpgBAshBOBzCBBe8CGIoF4oB4UA+ACjUxEKIEpcioUAoKFqRgEwgGMAbDeaADMArgDsuwAJYB7UVAC2CZOSwEAyvRIBnLQC4oa6vpWUAJABVhYHhBgYw+NQBpYS1OixEijRl1lbgKAwARn0AbVFheQAjBBcA+ElRRBdo6WkbDFEAXTDs3AU3ExIwsODslzCAcgALSSqXKuiQCCqKqDDgeGEIbOzqX39AjAAmcMiYuKgEpJSoNIyILNz8vEUkdwoS4Mra+sbm1vbO7t7+wdEAoIBmAvXlDxASsuO9hqgmlrbzv0vhgBZxlFYvB4l1ZisChhrkA).  Maybe it'll be helpful.

packages/core/type-definitions/combinator/merge.d.ts
TylorS: I think I'm in favor of adding TODO comments and waiting a few releases. Virtually *every* TypeScript release is a breaking change and updating to 3.4 syntax will break any users not yet on that version.

briancavalier: > Virtually every TypeScript release is a breaking change

😬 yikes!

Ok, sounds like we're all ok with adding TODOs for now.  I'll push an update with those soon, and then think about what we should do about documenting this in the face of the TS/Flow disparity.

TylorS: Yeah to clarify my previous statement - every `minor` release is potentially breaking. Every patch release in theory shouldn't be. I don't believe they follow semver in any regards TBH. I had a lot more run ins with that during the 2.x releases but I still don't really trust the version numbers

packages/core/src/index.js.flow
```diff
@@ -75,7 +75,9 @@ declare export function switchLatest <A> (s: Stream<Stream<A>>): Stream<A>
 
 declare export function merge <A, B> (s1: Stream<A>, s2: Stream<B>): Stream<A | B>
 declare export function merge <A, B> (s1: Stream<A>): (s2: Stream<B>) => Stream<A | B>
-declare export function mergeArray <A> (ss: Array<Stream<A>>): Stream<A>
+
+type MergeArray = <A>(Array<Stream<A>>) => Stream<A>
+declare export function mergeArray <S> (ss: S): $Call<MergeArray, S>
```
briancavalier: @unscriptable This isn't quite the same as the ZipArray case (which I still don't know how to do in Flow), but it's a useful technique nonetheless.  It basically turns an array type into a union of the array's value types (removing the extra Stream nesting in this case).

briancavalier:
> Figure out the implications for docs, given that Flow can't handle this (afaict)

After looking at the docs, I'm inclined just to go with them as is, and let the Flow limitations manifest themselves as type errors for now.  IOW, let's see if the current Flow arity limitations are problem _in reality_ before we try to figure out what to do about it.

packages/core/src/index.js.flow
```diff
@@ -75,7 +75,9 @@ declare export function switchLatest <A> (s: Stream<Stream<A>>): Stream<A>
 
 declare export function merge <A, B> (s1: Stream<A>, s2: Stream<B>): Stream<A | B>
 declare export function merge <A, B> (s1: Stream<A>): (s2: Stream<B>) => Stream<A | B>
-declare export function mergeArray <A> (ss: Array<Stream<A>>): Stream<A>
+
+type MergeArray = <A>(Array<Stream<A>>) => Stream<A>
```
Frikki: Hmmm ... shouldn’t this be:

```typescript
type MergeArray = <A>(ss: Array<Stream<A>>) => Stream<A>
```

And why not `ReadonlyArray<Stream<A>>`?

briancavalier: Flow doesn't need the parameter names in function types, only the parameter types.  So, the `ss:` isn't necessary.  Historically, if I remember correctly, we created this flow type declaration file around the time when flow had recently made that change (not requiring parameter names), so we included them for compatibility with folks who might be using previous versions.  At this point, it's safe (it's been years).

It's probably safe to use [Flow's `$ReadonlyArray`](https://flow.org/en/docs/types/arrays/#toc-readonlyarray) at this point, as well.   Thanks 👍 I'll make that change.

Issue #284, opened 2019-05-13
Duplicate addition of offset time to local time on `switch`?


gitusp:
On https://github.com/mostjs/core/blob/master/packages/core/src/combinator/switch.js#L113, `this.min` seems to end up being added to inner's local time twice when the inner `Segment` receives `end` or `error`.

---

At first, `this.min` is added to its local time on:
- https://github.com/mostjs/core/blob/master/packages/core/src/combinator/switch.js#L105
- or
- https://github.com/mostjs/core/blob/master/packages/core/src/combinator/switch.js#L109

Then, the method calls outer's `_disposeInner`, and it eventually goes into inner's `_dispose`, which adds `this.min` to local time again.

I'm sorry that I have not tested it.
But if the concern is true for you too, I'll write a test and open a PR for it.
Thank you.

briancavalier:
Hi @gitusp.  Thanks for opening an issue for this.  It does indeed look like a bug, and we'd very much appreciate a PR, especially one that starts with a failing test, as you mentioned.

Thanks!

gitusp:
@briancavalier Thank you for checking out the issue. I'll give a shot soon.

PR #285, opened 2019-05-19
Fix duplicate addition of offset time on switchLatest

gitusp:
This fixes https://github.com/mostjs/core/issues/284.

I confirmed that the original `switchLatest` emits an error event with time `2`, which should be `1`, and removing ` + this.min` from `Segment#_dispose` resolves the issue.

To be honest, I find my way of writing test case pretty clumsy, but I could not find a  straightforward way to examine the case.
Feel free to request changes, I'll appreciate it.

Thank you.

### DIFF
--- a/packages/core/src/combinator/switch.js
+++ b/packages/core/src/combinator/switch.js
@@ -110,6 +110,6 @@ class Segment {
   }
 
   _dispose (t) {
-    tryDispose(t + this.min, this.disposable, this.sink)
+    tryDispose(t, this.disposable, this.sink)
   }
 }
--- a/packages/core/test/switch-test.js
+++ b/packages/core/test/switch-test.js
@@ -11,7 +11,7 @@ import { ticks, collectEventsFor, makeEvents, makeEventsFromArray } from './help
 import { runEffects } from '../src/runEffects'
 
 describe('switch', () => {
-  it('given c anonical empty string, should return canonical empty', () => {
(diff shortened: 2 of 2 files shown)


briancavalier:
Hey @gitusp, thanks for doing this!

This is very much an edge case.  That said, I agree it's probably harder than it should be to observe the time value in this particular case.

Here's a thought on a slightly simpler approach using some of the testing utilities in core/test/helpers.  I'm sorry these are as well-documented as we'd like them to be.  FakeDisposeStream is from test/helpers/FakeDisposeStream.

```js
// new imports
import FakeDisposeStream from './helper/FakeDisposeStream'
import { run } from '../src/run'

 // ...
      const inner = FakeDisposeStream.from(() => { throw new Error() }, at(1))
      const s = at(1, inner)
      const sink = {
        event () {},
        end (t) {},
        error (t, e) {
          try {
            eq(2, t)
            done()
          } catch (e) {
            done(e)
          }
        }
      }

      run(sink, ticks(2), switchLatest(s))
```

What do you think?

gitusp:
Thank you @briancavalier! That's way simpler and cleaner.

I applied your suggestion and pushed again.
Sorry for having you write almost all the code. 🙇

briancavalier:
> Sorry for having you write almost all the code.

Haha, no worries at all, @gitusp.  Team work!

briancavalier:
Published in `@most/core` 1.3.5.  Thanks again!

PR #543, opened 2020-10-21
zipArray, combineArray - Support Literal & Dynamic array type

nissoh:
Hey everyone

Currently, we don't support either dynamic or literal types. current version of `@most/core 1.6.1` outputs the following error for both usages

```typescript
const newLocal = [now(22), now('ee')]
const combineTest = combineArray((scale, drag) => ({ transform: `scale(${scale})` }), newLocal);

/**
Argument of type '(Stream<string> | Stream<number> | Stream<null>)[]' is not assignable to parameter of type 'readonly [a: Stream<string | number | null>, b: Stream<string | number | null>]'.
  Target requires 2 element(s) but source may have fewer.ts(2345)
*/

````

I've made the following tests in both versions of typescript `^3.5.2` and `4.0.3`, they seem to output the same results
``` typescript
const dynamic = [now(1), now('a'), now(null)]                    // (Stream<number> | Stream<string> | Stream<null>)[]
const readonlyLiteral = [now(1), now('a'), now(null)] as const   // readonly [Stream<number>, Stream<string>, Stream<null>]
const literal = [now(1), now('a'), now(null)] as [Stream<number>, Stream<string>, Stream<null>]

combineArray((a, b) => a, dynamic) // a: string | number | null, b: string | number | null
combineArray((a, b) => a, readonlyLiteral) // a: number, b: string 
combineArray((a, b) => a, literal) // a: number, b: string 
combineArray((a) => a, []) // a: unknown

zipArray((a, b) => null, dynamic) // a: string | number | null, b: string | number | null
zipArray((a, b) => null, readonlyLiteral) // a: number, b: string 
zipArray((a, b) => null, literal) // a: number, b: string 
zipArray((a) => null, []) // a: unknown

```

I hope this resolves all use cases, please check PR and let me know if i didn't miss anything

### DIFF
--- a/packages/core/src/combinator/combine.ts
+++ b/packages/core/src/combinator/combine.ts
@@ -7,7 +7,7 @@ import { IndexSink, IndexedValue } from '../sink/IndexSink'
 import { disposeAll, tryDispose } from '@most/disposable'
 import invoke from '../invoke'
 import { Stream, Sink, Scheduler, Disposable, Time } from '@most/types'
-import { ToStreamsArray } from './variadic'
+import { InputStreamArray } from './variadic'
 
 /**
  * Combine latest events from two streams
@@ -27,16 +27,16 @@ export const combine = <A, B, C>(f: (a: A, b: B) => C, stream1: Stream<A>, strea
 * @returns stream containing the result of applying f to the most recent
 *  event of each input stream, whenever a new event arrives on any stream.
 */
-export const combineArray = <Args extends unknown[], R>(f: (...args: Args) => R, streams: ToStreamsArray<Args>): Stream<R> =>
+export const combineArray = <Args extends unknown[], R>(f: (...args: Args) => R, streams: InputStreamArray<Args>): Stream<R> =>
   streams.length === 0 || containsCanonicalEmpty(streams) ? empty()
     : streams.length === 1 ? map(f as any, streams[0])
-      : new Combine(f, streams)
+      : new Combine(f, streams as Stream<Args>[])
 
 class Combine<Args extends unknown[], B> implements Stream<B> {
   private readonly f: (...args: Args) => B
-  private readonly sources: ToStreamsArray<Args>;
+  private readonly sources: Stream<Args>[];
 
-  constructor(f: (...args: Args) => B, sources: ToStreamsArray<Args>) {
+  constructor(f: (...args: Args) => B, sources: Stream<Args>[]) {
     this.f = f
     this.sources = sources
   }
--- a/packages/core/src/combinator/variadic.ts
+++ b/packages/core/src/combinator/variadic.ts
@@ -1,9 +1,7 @@
 import { Stream } from '@most/types'
 
-// Map arrays to arrays of Streams:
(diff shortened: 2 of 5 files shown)


packages/core/src/combinator/combine.ts
```diff
@@ -27,16 +27,16 @@ export const combine = <A, B, C>(f: (a: A, b: B) => C, stream1: Stream<A>, strea
 * @returns stream containing the result of applying f to the most recent
 *  event of each input stream, whenever a new event arrives on any stream.
 */
-export const combineArray = <Args extends unknown[], R>(f: (...args: Args) => R, streams: ToStreamsArray<Args>): Stream<R> =>
+export const combineArray = <Args extends unknown[], R>(f: (...args: Args) => R, streams: InputStreamArray<Args>): Stream<R> =>
   streams.length === 0 || containsCanonicalEmpty(streams) ? empty()
     : streams.length === 1 ? map(f as any, streams[0])
-      : new Combine(f, streams)
+      : new Combine(f, streams as Stream<Args>[])
```
Frikki, line 33: Hmm!  I am not so happy about this casting because it means we have misaligned types.

Frikki (changes requested):
I have a slight concern about the typecasting. I think it needs some work before merging.

packages/core/src/combinator/combine.ts
nissoh: hey @Frikki , i got rid of the casting, there are still room for improvements in both versions(PR and master), both are not strictly correct, perhaps we can improvement it when typescript gets upgraded in the future.

briancavalier:
Thanks for putting this together, @nissoh.  Sorry I haven't been able to take a look yet.  I'm definitely in favor of this direction.  I'll take a closer look this week.

packages/core/src/combinator/variadic.ts
```diff
@@ -1,9 +1,7 @@
...
-// Map arrays to arrays of Streams:
-// Array<A> => Array<Stream<A>>
-// [A, B, C, ...] => [Stream<A>, Stream<B>, Stream<C>, ...]
-// TODO: use readonly any[] once TS 3.4.x has been in the wild for "enough" time
-export type ToStreamsArray<A extends ArrayLike<any>> = {
+type LiteralArray<A extends ArrayLike<any>> = {
   [K in keyof A]: Stream<A[K]>
 }
+
+export type InputStreamArray<Args extends ArrayLike<any>> = Readonly<LiteralArray<Args> | Stream<unknown>[]>
```
briancavalier, line 7: @nissoh I'm not sure I understand why we need `| Stream<unknown>[]` here.  Can you say a bit more about the use cases it helps?  Thanks!

nissoh: `| Stream<unknown>[]`, serve as compatibility to dynamic arrays. without it, dynamic input of dynamic arrays throws type error

There might be a better and understandable solution with typescript version 4

Frikki: Do we have a test for dynamic arrays?

PR #571, opened 2021-01-21
[Maybe Bugfix] Change awaitPromises/fromPromise not to emit events after disposed

gitusp:
Hi! It's been a while since I last made a little contribution to this project.
I'm very surprised that this whole project migrated to TypeScript, which is great! 🎉 

---

I found an edge case where awaitPromises/fromPromise unexpectedly emit events after disposed.
The case may occur when you use switchLatest with slow resolving Promise.

I added a test case, which fails before this modification, and passes after this.
I find managing `disposed` flag inside AwaitSink a bit awkward... but could not make it in another way, since Promise is not cancellable.

---

Anyway, thanks for maintaining this precious project. 🙏

### DIFF
--- a/packages/core/src/combinator/promises.ts
+++ b/packages/core/src/combinator/promises.ts
@@ -2,6 +2,7 @@
 /** @author Brian Cavalier */
 /** @author John Hann */
 
+import { disposeBoth } from '@most/disposable'
 import fatal from '../fatalError'
 import { now } from '../source/now'
 import { empty, isCanonicalEmpty } from '../source/empty'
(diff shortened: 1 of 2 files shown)


gitusp:
@briancavalier @TylorS _Just a friendly reminder_
I wonder if the behavior of current implementation is intended or not.
Could you give me a feedback at your earliest convenience?

Frikki (commented):
I’m trying to understand the test. I think a couple of comments showing intent on L91 would help.

gitusp:
@Frikki Thanks for reviewing! I added comments on the test case.

Frikki (approved):

TylorS:
Hey @gitsup, sorry for the long delay! @Frikki and I were able to chat today, and we're in agreement that there's a bug here that you've fixed. We realized we need to maintain our dependencies more actively as there's unrelated errors in our `test-latest` test cases. We're probably going to disable it, get a few PRs through the pipeline (this one included), address our dependencies issue, re-enable the tests and then do a release.

PR #673, opened 2024-02-07
adds 'core/last' emitting the last event. Useful for reduce/fold too.

semmel:
I sometimes need the last value in a stream or `reduce` it to the last value of `scan`

I could not find a way to do that with the existing API.

What do you think?

Also adds (helper function) `prelude/flow` to support function pipelines (in tests)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ jobs:
           name: Install Local JavaScript Dependencies
           command: |
             nvm use default
-            npm install
+            npm install --legacy-peer-deps
 
       # Store a cache of local JavaScript modules.
       - save_cache:
@@ -111,7 +111,7 @@ jobs:
           name: Install Local JavaScript Dependencies
           command: |
             nvm use node
-            npm install
+            npm install --legacy-peer-deps
 
       # Store a cache of local JavaScript modules.
       - save_cache:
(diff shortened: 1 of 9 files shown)


semmel:
I guess, Circle CI fails because one needs to `npm install --legacy-peer-deps` to install the project. `npm install` no longer works for me (npm v9.8).

TylorS:
We'll have to look into updating the build (hasn't been a need for a long time), but for this operator specifically, there's already a 3rd party library for this - 

https://github.com/mostjs-community/last

TylorS:
We do really appreciate the effort, and apologize if it feels like a waste of your time. If you'd we so kind to separate the CI fixes, I'd most graciously accept those.

If you're looking for method extensions to Stream, you might find this useful for inspiration at least https://github.com/mostjs/x-fluent. We've avoided fluent APIs in core because they can't be removed during tree-shaking, but they definitely are convenient

briancavalier:
Hi @semmel.  I'll echo @TylorS's apology for the effort you put into your `last` implementation.  Thanks for letting us know about `--legacy-peer-deps`, and for attempting FL support.  I'll take a look at https://github.com/mostjs/core/pull/675.

> That's a whole other story, but I hate to repeat myself in each file's imports

Using `import * as M from '@most/core'` is another option, if you don't mind namespace-style import.

Issue #239, opened 2018-05-17
Add Fantasy Land support


gabejohnson:
Continuing the conversation from https://github.com/cujojs/most/pull/347

briancavalier:
Hey @gabejohnson, thanks for opening this.  I've been watching https://github.com/fantasyland/fantasy-land/pull/286 closely.  It'd currently be much simpler for `@most/core` to implement [static-land](https://github.com/rpominov/static-land) (it mostly already does, but `of` is named `now`) than fantasy-land 3.x, since `@most/core` streams don't share a common prototype (they only share [an interface](https://github.com/mostjs/core/blob/master/packages/types/index.js.flow#L13)).

Hence my interest in https://github.com/fantasyland/fantasy-land/pull/286.  We'll continue keeping an eye on it.

Frikki:
And this https://github.com/fantasyland/fantasy-land/pull/315

semmel:
I've toyed around with two approaches integrating Fantasyland support. Both are backwards compatible. I guess, I'd favour the separate package approach (A), since it's not invasive.

### Benefits of not going for an add-on

- TypeScript throughout, (i.e. without importing compiled code)
- just a single compile-to-JS step

### Approach A) Provide a FL flavoured API in @most/core-fl by *wrapping* `Stream` with `class FantasyLandStream <A>` 

- provide a `FL<A>` Stream class which delegates FL method calls to @most/core functions. See [discussion][disc-1] and [playground code][playground]
- using the wrapper function, export *each* @most/core function as FL-flavoured API 

[Strawman PR](#675)

```typescript
// FL-API.ts
import FantasyLandStream from 'FLDelegate'
import { periodic as _periodic, take as _take, … } from '../core/src/index'
import { compose, curry2, curry3 } from '@most/prelude'
export const fantasyLand = <A>(stream: Stream<A>) =>
  new FantasyLandStream(stream);
  
// Number -> FantasyLandStream ()
export const periodic = compose(fantasyLand, _periodic)

interface Tap {
  <A>(f: (a: A) => any, s: Stream<A>): FantasyLandStream<A>
  <A>(f: (a: A) => any): (s: Stream<A>) => FantasyLandStream<A>
}
export const tap: Tap = curry2((x, y) => fantasyLand(_tap(x, y)))
// …
```

```typescript
// FLDelegate.ts
export class FantasyLandStream<A> implements Stream<A>, FunctorFantasyLand<A> {
[...]

PR #675, opened 2024-02-12
Strawman proposal for providing Fantasyland Stream type in a separate package

semmel:
This is a proposal how one can provide @most/core-fl by rebuilding @most/core and wrapping the Stream type with a FantasylandStream type.

- no `import @most/core`, so everything remains in TypeScript,
- @most/core-fl is 100% backwards compatible to @most/core

The implementation is simple, however the typings get quite laborious

See [Fantasyland support issue](https://github.com/mostjs/core/issues/239)

### DIFF
--- a/packages/core-fl/src/FantasyLandStream.ts
+++ b/packages/core-fl/src/FantasyLandStream.ts
@@ -1,0 +1,58 @@
+import { Stream, Sink, Scheduler, Disposable } from '@most/types'
+import {
+  continueWith,
+  empty,
+  filter,
+  map,
+  ap,
+  now,
+  chain,
+  never
+} from '../../core'
+
+interface FunctorFantasyLand<A> {
+  ['fantasy-land/map']<B>(fn: (a: A) => B): FunctorFantasyLand<B>
+}
+
+export const fantasyLand = <A>(stream: Stream<A>): FantasyLandStream<A> =>
+  new FantasyLandStream(stream)
+
+export class FantasyLandStream<A> implements Stream<A>, FunctorFantasyLand<A> {
+  constructor(private readonly stream: Stream<A>) {}
+
+  run(sink: Sink<A>, scheduler: Scheduler): Disposable {
+    return this.stream.run(sink, scheduler)
+  }
+
+  ['fantasy-land/concat'](nextStream: Stream<A>): FantasyLandStream<A> {
+    return fantasyLand<A>(continueWith(() => nextStream, this.stream))
+  }
+
+  ['fantasy-land/empty']<B>() {
+    return fantasyLand<B>(empty())
+  }
+
+  ['fantasy-land/filter'](predicate: (value: A) => boolean) {
+    return fantasyLand<A>(filter(predicate, this.stream))
+  }
+
+  ['fantasy-land/map']<B>(fn: (value: A) => B): FantasyLandStream<B> {
+    return fantasyLand<B>(map(fn, this.stream))
+  }
+
+  ['fantasy-land/ap']<B>(mapper: Stream<(value: A) => B>) {
+    return fantasyLand<B>(ap(mapper, this.stream))
+  }
+
+  ['fantasy-land/of']<B>(value: B) {
+    return fantasyLand<B>(now(value))
+  }
+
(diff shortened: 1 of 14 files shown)


packages/core-fl/src/FantasyLandStream.ts
```diff
@@ -0,0 +1,58 @@
...
+  new FantasyLandStream(stream)
+
+export class FantasyLandStream<A> implements Stream<A>, FunctorFantasyLand<A> {
+  constructor(private readonly stream: Stream<A>) {}
+
+  run(sink: Sink<A>, scheduler: Scheduler): Disposable {
+    return this.stream.run(sink, scheduler)
+  }
+
+  ['fantasy-land/concat'](nextStream: Stream<A>): FantasyLandStream<A> {
```
briancavalier, line 27: There's probably some discussion to be had on what a proper stream monoid should be. There might even be more than 1 valid monoid instance.

The way you've used `continueWith` here is nice.  It's like a simple `switch`.  I've come to believe that `merge` is also a pretty good semigroup.

Any thoughts or insights on which might be the better default semigroup instance?  Are there others?  See my other, related comment about `zero` as well.

packages/core-fl/src/FantasyLandStream.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  ['fantasy-land/of']<B>(value: B) {
+    return fantasyLand<B>(now(value))
+  }
+
+  ['fantasy-land/chain']<B>(fn: (value: A) => Stream<B>) {
+    return fantasyLand<B>(chain(fn, this.stream))
+  }
+
+  ['fantasy-land/zero']() {
```
briancavalier, line 55: The way I read the FL spec, `@more/core-fl` would also need to implement `alt`.  Is that right?

It seems like you'll need to ensure that `concat`+`empty` are coherent, as are `alt`+`zero`.  Have you thought about what makes sense for those in terms of coherence and their laws?

briancavalier (commented):
The overall approach here looks good, @semmel.  Thanks!

My current thinking is that this could be a separate package in [most-community](https://github.com/mostjs-community/about).  What do you think?

semmel:
> My current thinking is that this could be a separate package in [most-community](https://github.com/mostjs-community/about). What do you think?

Yes, of course. 

But is it smart to use TypeScript for that community package then? (Could be just a matter of personal taste though). For example I could rather implement the FL wrapper in JS: 
1. I'd `import {ap, continueWith, …} from '@most/core'` which is ES5 code generated from TypeScript.
2. in JS implement `fantasyland(stream)` returning a simple JS Object 
3. author TS typings in `index.d.ts` using '@most/types' *separate* from the JS implementation 
4. Then for my exports (nearly all of the @most/core API) I'd create a dist bundle which re-bundles to ES5?

I am not particular keen of working in TypeScript, but I value the type definitions.

Would it be better (e.g. aid debugging) if @most/core/dist/indes.es.js would be ES2020 and not ES5? 
That would be a breaking change of course.

briancavalier:
> Yes, of course.

Great!

> But is it smart to use TypeScript for that community package then? (Could be just a matter of personal taste though).

It's totally up to you.  You'll own the repo and the package, so if you prefer to build in JS and include .d.ts files for types (or not), that's totally cool with us 😄 . When you're ready, you can [open an issue here to request membership](https://github.com/mostjs-community/about/issues), and we'll add you as a collaborator.

> For example I could rather implement the FL wrapper in JS:

1-3 all sound fine to me.

> 4. Then for my exports (nearly all of the @most/core API) I'd create a dist bundle which re-bundles to ES5?

There might be other options, but I can think of two you could consider:

1. Make `@most/core` a dependency and bundle it with your package.
2. Make `@most/core` a peerDependency, not bundled with your package, and devs install it into their project separately.

I think 2 could be a bit more helpful for existing projects that are already using `@most/core` and want to start using it via your new fantasy-land package.

> Would it be better (e.g. aid debugging) if @most/core/dist/indes.es.js would be ES2020 and not ES5?

Yeah, you're right that targeting a new ES version is probably best.  I'm not sure when we'll get to it, but I agree it's a good idea.

PR #677, opened 2024-03-07
Fix: at now curried + some addition to documentation

### LINKED ISSUE #676: Is there a reason why `at(time, val)` is not curried?
Apart from `runEffects`, `at` seems to be the only (n>1)-ary function which is not auto-curried.

I find that surprising.
Is there a reason for it?

semmel:
fixes #676 

and docs `newStream` and adds PACKAGES.md to Readme.md

Just a quickly jotted down PR.

Happy to change/correct anything. 
:-)

### DIFF
--- a/PACKAGES.md
+++ b/PACKAGES.md
@@ -6,6 +6,7 @@
 | [hold](https://github.com/mostjs/hold) | Deliver the most recently seen event to new observers
 | [create](https://github.com/mostjs/create) | Imperatively push events into a Stream
 | [dom-event](https://github.com/mostjs/dom-event) | Streamlined DOM Events
+| [adapter](https://github.com/mostjs/adapter) | Imperatively put events into mostjs streams
 | [examples](https://github.com/mostjs/examples) | Most.js examples
 
 ### Most Tools for  Contributors
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -286,7 +286,8 @@ at
 
 .. code-block:: haskell
 
-  at :: Time -> a -> Stream a
+  at :: (Time, a) -> Stream a
+  at :: Time -> Stream void
 
 Create a :ref:`Stream` containing a single event at a specific time. ::
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -318,6 +318,34 @@ Create a :ref:`Stream` that fails with the provided ``Error`` at time 0.  This c
 
   throwError(X): X
 
+.. _newStream:
+
+newStream
+`````````
+
+.. code-block:: haskell
+
+  newStream :: ((Sink a, Scheduler) -> Disposable) -> Stream a
+
+Create a :ref:`Stream` from a custom event producer (i.e. :ref:`run` method).
+
+.. code-block:: javascript
+
+  import { currentTime as ct } from '@most/scheduler'
+
+  // Promise a -> Stream a
+  const fromPromise = p => newStream(
+    (sink, scheduler) => {
+      p.then(
+        x => sink.event(ct(scheduler), x),
+        e => sink.error(ct(scheduler), e)
+      )
+      .finally(() => sink.end(ct(scheduler));
+
+      return { dispose: () => undefined };
+    }
+  )
+
 Extending
 ^^^^^^^^^
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -333,18 +333,19 @@ Create a :ref:`Stream` from a custom event producer (i.e. :ref:`run` method).
 
   import { currentTime as ct } from '@most/scheduler'
 
-  // Promise a -> Stream a
-  const fromPromise = p => newStream(
+  // Number -> a -> Stream a
+  const valueAt = delay => value => newStream(
     (sink, scheduler) => {
-      p.then(
-        x => sink.event(ct(scheduler), x),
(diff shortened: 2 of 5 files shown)


PACKAGES.md
```diff
@@ -6,6 +6,7 @@
 | [hold](https://github.com/mostjs/hold) | Deliver the most recently seen event to new observers
 | [create](https://github.com/mostjs/create) | Imperatively push events into a Stream
 | [dom-event](https://github.com/mostjs/dom-event) | Streamlined DOM Events
+| [adapter](https://github.com/mostjs/adapter) | Imperatively put events into mostjs streams
```
briancavalier, line 9: 👍  Thanks

docs/api.rst
```diff
@@ -318,6 +318,34 @@ Create a :ref:`Stream` that fails with the provided ``Error`` at time 0.  This c
...
+  // Promise a -> Stream a
+  const fromPromise = p => newStream(
+    (sink, scheduler) => {
+      p.then(
+        x => sink.event(ct(scheduler), x),
+        e => sink.error(ct(scheduler), e)
+      )
+      .finally(() => sink.end(ct(scheduler));
+
+      return { dispose: () => undefined };
```
briancavalier, line 345: It'd be great to show a real disposable here, e.g. one that prevents the call to sink.event/error/end above.

packages/core/src/index.ts
```diff
@@ -2,15 +2,25 @@
...
-
 export { newStream } from './source/newStream'
 
+// -------------------------------------------------------
+
+import { at as _at } from './source/at'
+
+interface At {
+  <A>(time: Time, value: A): Stream<A>
+  <A>(time: Time): (value: A) => Stream<A>
```
briancavalier, line 20: Oh jeez, I'm sorry.  It didn't occur to me in https://github.com/mostjs/core/issues/676 that this might be a breaking change for _JavaScript_ users.  I just noticed it now that I see these type signatures.

In the non-curried version, `at`'s TS type requires the second parameter, i.e. the value, but a JS caller _could_ omit it, e.g. `at(time)`, and it would return a `Stream`.

By currying it, the JS behavior will change: instead of `at(time)` returning `Stream`, it will return a function `(value) => Stream` 😕 

We'd need to release it as 2.0.0 to signal the breaking change.  It's not a blocker, though, since I don't really care about the "marketing value" of version numbers.  I figured it was worth calling out, though.

That might be reason enough to separate this into 2 PRs: one for the non-breaking changes, and one for the breaking change to `at`.

Thoughts on all of that?

packages/core/test/source/at-test.ts
```diff
@@ -12,4 +12,11 @@ describe('at', () => {
     return collectEventsFor(t, at(t, expected))
       .then(eq([{ time: t, value: expected }]))
   })
+  it('is auto-curried', function () {
```
briancavalier, line 15: It's ok not to test this.  The types already guarantee it.  For example, here's the type error we'd get if we wrote this same code without at being curried:

Screen Shot 2024-03-07 at 6 56 02 PM

docs/api.rst
semmel: I have changed the example using setTimeout now with dispose.

packages/core/src/index.ts
semmel: Ok I understand, I too missed that this is a breaking change.

I think, currying `at` is a too small change that it alone not justifies releasing a "2.0". 
I reverted that commit, so that *this PR* now only deals with the non-breaking changes to the documentation.

I could author another PR for `at`, but rather I'd somehow bookmark the commit as "2.0-planned" to include later if the real need for a version "2.0" should arise. 

Perhaps you could mark the issue #676 with a label like "2.0"?

semmel: I've updated the signature for at in the documentation to reflect it's signature as uncurried binary and unary function.

briancavalier: > Perhaps you could mark the issue https://github.com/mostjs/core/issues/676 with a label like "2.0"?

Great idea!  Done.

docs/api.rst
```diff
@@ -286,7 +286,8 @@ at
 
 .. code-block:: haskell
 
-  at :: Time -> a -> Stream a
+  at :: (Time, a) -> Stream a
+  at :: Time -> Stream void
```
briancavalier, line 290: Unfortunately, the second signature is only valid for JavaScript.  The TS types currently prevent calling `at` with just 1 argument.

Maybe we can keep only the first signature, and add a note for JavaScript users?

```suggestion
```

docs/api.rst
```diff
@@ -318,6 +319,35 @@ Create a :ref:`Stream` that fails with the provided ``Error`` at time 0.  This c
...
+
+  // Number -> a -> Stream a
+  const valueAt = delay => value => newStream(
+    (sink, scheduler) => {
+      const publishVal = val => {
+        sink.event(ct(scheduler), val);
+        sink.end(ct(scheduler));
+      };
+      const timer = setTimeout(publishVal, delay, value);
+      return { dispose: () => { clearTimeout(timer); } };
```
briancavalier, line 345: 👍

briancavalier (approved):
Looks good.  Thank you!

